PR #8, opened 2018-02-26
PR now structure and approach somewhat stable 

cefn:
This PR is based on recently-tested and seemingly well-functioning Blinka compatibility layer, making it possible to use unmodified Adafruit_CircuitPython_XX modules on Micropython 1.9.3 on Feather Huzzah and Pyboard as per https://github.com/adafruit/Adafruit_Micropython_Blinka/issues/2 assuming they rely on...
* board
* microcontroller
* digitalio
* bitbangio.I2C
* busio.I2C
* busio.UART
* time

Happy to simplify if all these commits drive people up the wall. Future commits are expected to be feature-oriented, not line-by-line stream-of-consciousness now that the structure is settled.

Some intro documentation is at https://github.com/cefn/Adafruit_Micropython_Blinka/blob/master/examples/index.md but more documentation would be needed for adoption in the wild once other priorities are met.

I have skipped over busio.SPI and bitbangio.SPI testing for now, since the only reference SPI example from the Tasklist relied on the adafruit_register library (requires descriptor support within the build as per... https://github.com/adafruit/Adafruit_Micropython_Blinka/issues/6 ). 

However, there is an implementation of SPI which looks OK but as yet untested. I am ready to take your steer on an alternative SPI test you think would be representative, (I am awaiting the BNC cables and practical skills to be able to use my oscilloscope to judge SPI performance directly :)

Also awaiting instruction if a property-based monkey-patching of the adafruit_register descriptor-based classes is worth doing, which would mean I could tackle the RFM69 test directly based on the CircuitPython library. I think it might be feasible to achieve compatibility in this way - by wiring up the __get__ and __set__ dynamically using properties (which ARE available in Micropython) after library loading. However, I won't embark on this unless it's seen to be worthwhile, depending what level of seamless support the Blinka layer should have.

I will update @ladyada about run-rate (how many hours this is all taking) so it can be decided what to do next.


src/board.py
```diff
@@ -20,15 +20,25 @@
...
+platform introspection
 
 * Author(s): cefn
 """
 
-# imports
+import gc
+gc.collect()
+from adafruit_blinka.agnostic import board
+gc.collect()
```
tannewt: Why are these gc.collects here?

cefn:
In earlier testing, I was frequently hitting memory ceiling (on Feather M0 Express I think) but just needed a few extra bytes. The gc.collect() in between imports were an attempt to minimise heap fragmentation, given Micropython doesn't have allocation from opposite ends of the memory space to assist here, at least not yet.

PR #20, opened 2018-07-24
Write_ReadInto PR Review Changes

brentru:
**Adding requested changes from previous pr:**

 
> runtime error should have a message explaining what went wrong

`raise RuntimeError('Buffer slices must be of equal length.')`

> checks if in_end/out_end is none and have them set to the 'default' len(array)

```
        if out_end is None or in_end is None:
            out_end = len(buffer_out)
            in_end = len(buffer_in)
```

>  use the out_start & out_end ranges here to limit what is sent

`            data = self._spi.xfer(list(buffer_out[out_start:out_end]))
`

> use the in_start & in_end ranges here to put the data in the right section of the buffer_in array

`            for i in range(len(buffer_in[in_start:in_end])): # 'readinto' the given buffer
`

### DIFF
--- a/src/adafruit_blinka/microcontroller/raspi_23/spi.py
+++ b/src/adafruit_blinka/microcontroller/raspi_23/spi.py
@@ -71,8 +71,11 @@ class SPI:
     def write_readinto(self, buffer_out, buffer_in, out_start=0, out_end=None, in_start=0, in_end=None):
         if not buffer_out or not buffer_in:
             return
+        if out_end is None or in_end is None:
+            out_end = len(buffer_out)
+            in_end = len(buffer_in)
         if out_end - out_start != in_end - in_start:
-            raise RuntimeError
+            raise RuntimeError('Buffer slices must be of equal length.')
         try:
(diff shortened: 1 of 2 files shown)


src/adafruit_blinka/microcontroller/raspi_23/spi.py
```diff
@@ -82,8 +85,8 @@ def write_readinto(self, buffer_out, buffer_in, out_start=0, out_end=None, in_st
             self._spi.max_speed_hz = self.baudrate
             self._spi.mode = self.mode
             self._spi.bits_per_word = self.bits
-            data = self._spi.xfer(list(buffer_out))
-            for i in range(len(buffer_in)): # 'readinto' the given buffer
+            data = self._spi.xfer(list(buffer_out[out_start:out_end]))
+            for i in range(len(buffer_in[in_start:in_end])): # 'readinto' the given buffer
```
ladyada: i dont think this section does what you think it does - please print/debug to make sure the data is going into the buffer_in slice!

PR #22, opened 2018-08-12
I2C and SPI fixes to bring more in line with busio

ladyada:
@kattni this will make LIS3DH work!


ladyada:
thanx, plz also do a release!

PR #32, opened 2018-08-31
add missing module

ladyada:


brentru:
Works with `> sudo python3` but _not_ as user `python3`. Looks like this is within rpi_ws281x and not blinka. 

`RuntimeError: ws2811_init failed with code -5 (mmap() failed)`

ladyada:
thats notrmal - you need root access to use the peripherals :(

PR #33, opened 2018-08-31
runtimeerror msg using neopixel_write without sudo 

brentru:
Calls to `neopixel_write` fail when python is not run as sudo with:
`RuntimeError: ws2811_init failed with code -5 (mmap() failed)`

Let's catch that error, send a message to users to alert them to prefix their script with `sudo`


brentru:
Ok - the init() call checks if the ws2811 was successful and raises a runtime error with a resp. (https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/raspi_23/neopixel.py#L52) 

Should I be checking for a `resp` of (-5), "mmap() failed"?

ladyada:
yah lets start there

ladyada:
i tweaked and put into https://github.com/adafruit/Adafruit_Blinka/pull/58

PR #34, opened 2018-09-08
only install on raspi?

ladyada:


dhalbert:
Note: this is an admittedly imperfect way of requiring `rpi_ws281x` only on rpi.

PR #47, opened 2018-11-23
add RGBW support by counting # of bytes in buffer

ladyada:


ladyada:
anyone wanna take a look :)

jerryneedell:
What I am asking is how do you distinguish 16 RGB from 12 RGBW?

ladyada:
you can't - but it doesnt matter, its sent out as one DMA chunk

ladyada:
thank ya :)

PR #48, opened 2018-11-26
Libgpiod pulsein prelim support

ladyada:
This relies on a precompiled & staticly linked libgpiod helper program which can live in the folder for now (long term we should have it auto-compile the latest version but thats a packaging adventure since libgpiod does not come with many distros)

implements PulseIn, slowly, but somewhat effectively. tested with DHT on raspi1 and pi 3. DHT library needs update to avoid contexts because the loading/unloading of the program takes a while - maybe better to just have one pulsein object for the life of the sensor. also DHT decoding could be improved, theres a lot of false errors.

i realy wanna get rid of the DHT Python library, its an install/maintenance frustration :)


brennen:
Note there might be license issues with distributing the `libgpiod_pulsein` binary this way.  libgpiod is LGPL, not full GPL, but we should make sure we're on the up with that.

ladyada:
huh really? i thought LGPL is less constrained, we'll can have a README saying where the code is

brennen:
Yeah, we might be fine; we are publishing the source after all.  I'm just not sure if it matters whether we statically link libgpiod instead of using it as a shared lib. I'll do a bit of reading just in case.

ladyada:
we're fine http://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic, all source is released

brennen:
cool.

ladyada:
zoom ok - im going to try this on an orangepi and if it works there by magic ill move to generic-linux

PR #51, opened 2018-11-26
renamed spi/i2c to generic linux since they're not platform specific

ladyada:
and updated busio to include it when possible


src/busio.py
```diff
@@ -16,10 +16,8 @@ def __init__(self, scl, sda, frequency=400000):
 
     def init(self, scl, sda, frequency):
         self.deinit()
-        if board_id == "raspi_3" or board_id == "raspi_2":
-            from adafruit_blinka.microcontroller.raspi_23.i2c import I2C as _I2C
-        if board_id == "beaglebone_black":
-            from adafruit_blinka.microcontroller.raspi_23.i2c import I2C as _I2C
+        if board_id == "raspi_3" or board_id == "raspi_2" or board_id == "beaglebone_black":
```
caternuson: How about stashing all these in a tuple and then doing something like:
```python
if board_id in LINUX_BOARDS:
```

ladyada: ill do that long term but i need to redo the platform detection code furst

caternuson: coolio.

ladyada:
@caternuson all good by you?

caternuson:
Sry, yep. All good. I just had the one casual suggestion.

PR #54, opened 2018-11-27
add generic linux pin class using libgpiod python bindings

brennen:
**WORK IN PROGRESS, DO NOT MERGE**


src/adafruit_blinka/microcontroller/generic_linux/pin.py
```diff
@@ -0,0 +1,66 @@
...
+        return str(self.id)
+
+    def __eq__(self, other):
+        return self.id == other
+
+    def init(self, mode=IN, pull=None):
+        if mode != None:
+            if mode == self.IN:
+                flags = ()
+                if pull != None:
```
brennen: Is this section at all on the right track?

ladyada:
i think this looks good - just need to make sure it works :)

brennen:
Yeah, I've tested a little bit with just a button, but I was planning to plug it into the Pi class and check against some real code.

ladyada:
try gutting out the RPi.GPIO code we have now, see if it works

ladyada:
@brennen is this ready for testing/mergin?

ladyada:
i cherrypicked this for my orangepi port - turns out this isnt a good choice for raspi because it doesnt have pullup support. so we'll stick with rpi.gpio for now. closing but your code lives on https://github.com/adafruit/Adafruit_Blinka/pull/59 !!!

PR #55, opened 2018-11-28
add RPi.GPIO to requirements

brennen:
We're looking to dispense with this, but in the meantime I think it's breaking a fair number of installs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 Adafruit-GPIO
+RPi.GPIO
 rpi_ws281x>=4.0.0; platform_machine=='armv7l'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 Adafruit-GPIO
-RPi.GPIO
-rpi_ws281x>=4.0.0; platform_machine=='armv7l'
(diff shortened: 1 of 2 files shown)


ladyada:
hmm i think this will break on non-Pi's right? can you put a warning on the import instead?

ladyada:
https://github.com/adafruit/Adafruit-Retrogame/blob/master/joyBonnet.py#L24

brennen:
I had that thought and tested on my desktop system - nothing seems to complain.  It presumably wouldn't _work_, but I don't think install is really platform limited?

ladyada:
no idea. i am very bad at python packaging :D

brennen:
With last couple of commits:  Probably Good Enough.

ladyada:
seems good to me @dhalbert any thoughts?

dhalbert:
If those changes work, then OK. I would suggest not suggesting `sudo pip` in general in those error messages. It's a bad idea because it can conflict with distribution-supplied packages.

brennen:
@dhalbert yeah, going the route of requiring RPi.GPIO on arm6/7 vs. giving install instrux in an error message.  Should fix the problem and even if that's casting a bit of a broad net, I don't think having it installed on non-Pi devices does any harm.

PR #58, opened 2018-12-01
RGBW support for NeoPixels

ladyada:


ladyada:
yah that would require making bpp global right? i'd like to avoid em because we may eventually support multiple strips, this way we query the strip object directly

Issue #62, opened 2018-12-08
Bus IO and UART definitions in processor file rather than in board file


Tonymac32:
For the experimental support of H3 (and so H2+ and H5) devices, the bus peripherals and UARTS can be anything from the SoC, and there are numerous non-pi form factor boards.  It makes more sense to me to set the UART/SPI/I2C in the specific board files, leaving the pin definition to define only the pin names.

ladyada:
its something we thought about but here's our reasoning:
each *chip* has specific pins that can do UART/I2C/SPI, etc. and those are fixed capabilities. each board may or may not have the DTO for that peripheral. we don't want to do DTO loading within Blinka, and people can change around the DTO. for example in raspi, there's a 2nd hardwaer SPI port, that most people don't enable, but we still have support for it in case they do. we use the special names in the board file 'SDA' 'MOSI' to hint the most popular hardware setups

thoughts?

Tonymac32:
I agree that each chip has it's specified I/O, however they are almost always multiplexed and which of those is being used by a vendor or in a default Armbian configuration on a specific board is always different, The Tritium is using UART-1 on pins 8 and 10, the Orange Pi PC is UART-3 in that position, etc. 

You get the same situation with Amlogic, the C2 and the "Le Potato" exposes UARTA on pins 8 and 10.  NanoPi K2 exposes UARTC.  Do you put both in the chip definition?  And if you do, can those pins be used as normal GPIO at the same time? I haven't traced all the code, I'm just assuming defining the UART results in it being assigned a fixed function, meaning, to make sure the pin 8 and 10 uart is available on every pi-compatible allwinner board will result in boards like Tritium having 4 GPIO's tied up and blocking SPI1.

ladyada:
ok now that i think of it - UART isnt actually something we'll be exposing thru blinka because we recommend people use `pyserial` because the names change a lot, even between distros. how about i2c/spi?

ladyada:
yeah its one of those "ugh which decision is right? no way to know and both have downsides" type things :) we'll be thinking about it, part of adding support for a lot of linux boards is we're expecting to bump into all sorts of edge cases!

Tonymac32:
well, that's sort of why I was asking this question:  If you define all the pin names in the _chip_ file, then define the "default" ports in the _board_ file, each board can tell the library what is the preferred spi/i2c/etc.  Then some kind of function to spit out "available resources".

[...]

PR #63, opened 2018-12-08
Add board support files for Tritium H3

Tonymac32:
This adds only the board file and extra needed pins to support the Libre Computer Tritium family: https://libre.computer/products/boards/all-h3-cc/ .  Obviously running on the board will not work with this patch alone, the various detects need put in place.

As the board detection is in flux I haven't pushed that part, and the UART port defined by the H3 file is OPi-PC specific, but will work if you have the pin map of the Tritium board handy, mentioned in https://github.com/adafruit/Adafruit_Blinka/issues/62

Armbian Board Name "tritium-h3"

### DIFF
--- a/src/adafruit_blinka/board/tritium-h3.py
+++ b/src/adafruit_blinka/board/tritium-h3.py
@@ -1,0 +1,39 @@
+from adafruit_blinka.microcontroller.allwinner_h3 import pin
+
+PA12 = pin.PA12
+SDA = pin.PA12
+PA11 = pin.PA11
+SCL = pin.PA11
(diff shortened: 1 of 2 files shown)


ladyada:
awesome. thank you! we have the platform detection library here - can you try to add support for your board there (detection) - we will be merging that in to replace what we've got now which should simplfy adding boards :)

https://github.com/adafruit/Adafruit_Python_PlatformDetect

ladyada:
ok merged! please submit more PRs when the rest of the plumbing is done, should be much easier now :)

PR #67, opened 2018-12-28
pi 1 pin definitions

brennen:
Add correct pin definitions for two variants of Raspberry Pi 1.


brennen:
@ladyada added rev2.

ladyada:
kk is this ready to test?

brennen:
should be!

ladyada:
tested the v2, works!

schustykd:
could someone advise if 26 pin raspi_1a is supported by blinka.

detect.py correctly identifies my board as RASPBERRY_PI_A but blinkatest.py fails with "Board not supported".

any advice is appreciated. sorry if this inquiry is in the wrong place - i'm new.

ladyada:
please make a new issue!

PR #71, opened 2018-12-29
support 26 pin rpi

Dmole:


ladyada:
hiya thanks, this is going to be merged from https://github.com/adafruit/Adafruit_Blinka/pull/67 which differentiates the two 26-pin Pi 1's

PR #73, opened 2018-12-30
spidev only on linux (where it exists)

ladyada:
check and only try to install spidev on linux (keeps blinka from failing on mac/win)


brennen:
tested install on pi.  looks good.

PR #78, opened 2019-01-15
working libgpiod_pulsein on pi: add sysv_ipc; add pulseio to py_modules

brennen:
The included libgpiod_pulsein binary wasn't being used on the Pi, specifically with Adafruit_CircuitPython_DHT.  It turns out that `pulseio` was missing from the list of modules to install, we weren't installing `sysv_ipc` as a dependency, and `libgpiod_pulsein` itself wasn't marked as executable.

This is now tested and working on a Pi 3B+ with a DHT22.


brennen:
Whoops, didn't merge this after approval.  Release shortly.

cc: @PaintYourDragon

PR #80, opened 2019-01-25
first pass at adding correct i2c bus for original raspberry pi

brennen:
Per troubles here:

https://forums.adafruit.com/viewtopic.php?f=8&t=146375

It turns out the only Pi I have that's (maybe) old enough to reproduce this won't boot, but I _think_ this is what needs to happen, based on these lines:

https://github.com/adafruit/Adafruit_Blinka/blob/master/src/busio.py#L25

I have confirmed that this doesn't break anything on the next newest Pi model after the target one.

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -1,4 +1,6 @@
 import RPi.GPIO as GPIO
+from adafruit_blinka.agnostic import detector
+
 GPIO.setmode(GPIO.BCM)    # Use BCM pins D4 = GPIO #4
 GPIO.setwarnings(False)   # shh!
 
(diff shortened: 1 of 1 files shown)


src/adafruit_blinka/microcontroller/bcm283x/pin.py
```diff
@@ -116,7 +118,12 @@ def value(self, val=None):
     (1, TXD, RXD),
 )
 
-i2cPorts = (
-    (1, SCL, SDA),
-)
-
+# Set correct i2c bus - original Pi used 0, subsequent models use 1:
+if detector.board.RASPBERRY_PI_B_REV1:
```
brennen: It feels a bit weird to be doing this in the pin definition file for the bcm283x, but I wasn't quite sure how else to go about it.

ladyada:
editing into another PR

PR #82, opened 2019-01-27
Add secondary i2c pins

JustinCredible-:
Add additional i2c pins


caternuson:
Where did those specific pins come from? The entry should be added as another 3 tuple. See here for example:
https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/stm32/pin.py#L65 

This PR appears to be conjoined with #81.

PR #83, opened 2019-01-29
fix for first pi's

ladyada:
as seen here
https://forums.adafruit.com/viewtopic.php?f=8&t=146375&p=725923#p725923
i tested with hardware too

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -117,6 +117,6 @@ uartPorts = (
 )
 
 i2cPorts = (
-    (1, SCL, SDA),
+    (1, SCL, SDA), (0, D1, D0),   # both pi 1 and pi 2 i2c ports!
 )
 


src/adafruit_blinka/microcontroller/bcm283x/pin.py
```diff
@@ -117,6 +117,6 @@ def value(self, val=None):
 )
 
 i2cPorts = (
-    (1, SCL, SDA),
+    (1, SCL, SDA), (0, D1, D0),   # both pi 1 and pi 2 i2c ports!
```
caternuson: Put new entry on a second line? And could also be ([they are equiv](https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/board/raspi_1b_rev1.py#L7)):
```python
(0, SCL, SDA), # original Rev 1 boards (rare)
```
unless you want to use the `D0` and `D1` labels for differentiation.

ladyada: nope they're not - this has to be done this way - the SDA/SCL pin names are defined in the board, they will work the same on any board

caternuson: thanks. gotcha, see it now. it's the [microcontroller pins](https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/bcm283x/pin.py#L62) being used at this point.

for the code comment - do you mean "both rev 1 and rev 2"? as is, sounds like pi 3 is left out

ladyada: if we can merge this - and then fix commentary later - i need to keep movin'

caternuson: sure thing.

ladyada: thanks :)

PR #88, opened 2019-03-08
Jetson Integration

atan1021:


src/adafruit_blinka/microcontroller/tegra/pin.py
```diff
@@ -0,0 +1,132 @@
+import sys
+import Jetson.GPIO as GPIO
+sys.path.append("/opt/nvidia/jetson-gpio/lib/python")
```
ladyada: this is a little unusual - why is it here? you've already imported the GPIO library?

src/adafruit_blinka/board/jetson_pin.py
```diff
@@ -0,0 +1,48 @@
+"""Pin definitions for Jetson models."""
+
+from adafruit_blinka.microcontroller.tegra import pin
+
+D1 = pin.D1
```
ladyada: please follow the naming conventions you have for the boards. e.g.
https://www.jetsonhacks.com/nvidia-jetson-tx2-j21-header-pinout/
`GPIO_GCLK` `GPIO5` rather than coming up with non-standard names that don't match the documentation.

there should be a `board` file for each different Jetson model, where the name `GPIO5` will match up to the chip's underlying true name, e.g. GPIO398

image

src/adafruit_blinka/microcontroller/tegra/pin.py
```diff
@@ -0,0 +1,132 @@
...
+sys.path.append("/opt/nvidia/jetson-gpio/lib/python/Jetson/GPIO")
+GPIO.setmode(GPIO.BCM)
+GPIO.setwarnings(False)   # shh!
+
+# Each Jetson model uses different I2C busses
+JETSON_I2C_BUS_DEFS = {
```
ladyada: this should go in each `board` file - doesnt belong here

src/adafruit_blinka/microcontroller/tegra/pin.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def cleanup(self, channel=None):
+        if channel is None:
+            GPIO.cleanup()
+        elif channel == self:
+            GPIO.cleanup(self.id)
+        else:
+            raise RuntimeError("Invalid pin to cleanup")
+
+D1 = Pin(1)
```
ladyada: this should be the actual name of the underlying Tegra microcontroller, see for example how we do it for allwinner:

https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/allwinner_h3/pin.py

`PG13 = Pin(205)` means GPIO205 (which is the libgpiod name) is also known as PG13 in the chip datasheet. basically, the names have to map to the underlying hardware not be named based on the location on a PCB connector

src/board.py
```diff
@@ -61,6 +61,9 @@
 elif board_id == ap_board.GIANT_BOARD:
     from adafruit_blinka.board.giantboard import *
 
+elif detector.board.any_jetson_board:
```
ladyada: you have to do it based on board_id, one for each board definition

src/digitalio.py
```diff
@@ -72,6 +74,12 @@ def switch_to_input(self, pull=None):
     def deinit(self):
         del self._pin
 
+    def cleanup(self, channel=None):
```
ladyada: this is not a `digitalio` standard interface function, so it can't go here. you can hvae an atexit() or something in your own Pin files, but it can't go here

src/adafruit_blinka/microcontroller/tegra/pin.py
atan1021: The Jetson.GPIO library already does the mapping. The library will not work if the names map to the tegra hardware directly
But I can change the naming of the pins to match the naming on the datasheet

ladyada: you can have a pin id if you like, but given you may change the GPIO library to use libgpiod some day, we should stay true to the hardware

atan1021: Unfortunately, the hardware GPIO cannot be directly accessed due to the nature of the Jetson.GPIO

Jetson.GPIO is designed to resemble RPi.GPIO so they are compatible. It either maps the physical BOARD pin #, BCM pin # as seen on the 40 pin RPi, CVM string name, or TEGRA_SOC string name to the hardware Linux GPIOs

Should the change be made to use libgpiod or update Jetson.GPIO mappings, another PR can be made to reflect this. Until then, there is no way to directly stay true to the hardware

ladyada: OK - please use native chip names in the Tegra **chip** definition, and BCM naming in the board definition, that will make it the most compatible with our existing Pi libraries and examples

atan1021:
Each Jetson platform has its own tegra chip with native TEGRA_SOC names associated with it that maps to the hardware. So I added chip definitions for each tegra chip.
The same goes for board definitions as you mentioned in the previous commit.

ladyada:
ok just to verify - does each board use a different tegra chip model number? like they're totally different chips?

atan1021:
yeah each board uses a different chip model
They are similar in ways but they do have different pinouts, which is what made the need for making multiple definitions The Jetson.GPIO library can abstract all of this (reducing it back to one file), but it would be going back to mapping the PCB connector which we don't want

ladyada:
ok - just to verify the names of the chip are literally "Tegra TX1" there is no other model #?

atan1021:
Oops the chip is "Tegra X1" but other model number is T210
The same goes for the other chips, they have an another model #

ladyada:
ok please name `chip` devices by the true hardware name :)

atan1021:
I realized that "TEGRA_TXX" also shares model number T210, but is a different version and also has a different pin configuration
"TEGRA_TXX" and "TEGRA_X1" are still technically the names of the chips though if that naming is fine

ladyada:
okidoke - so you need to update platformdetect?

atan1021:
yeah just need to refactor and test some things first!

atan1021:
Is this reorganization acceptable?

I also saw some whitespace/no new line issues in some of the files. I'll fix those on the next commit

ladyada:
lets do one release only - so clean up any whitespace/lines you want now

ladyada:
hiya i see a lot of `board_id` import additions, but they're not used? keep imports minimal please - if you dont use it, dont import it :)

atan1021:
> hiya i see a lot of `board_id` import additions, but they're not used? keep imports minimal please - if you dont use it, dont import it :)

I removed the import `board as ap_board` which isn't used, but the `board_id` I added is used

johnpanos:
i love you guys thanks

ladyada:
you shouldn't be referencing the `board` type then, if t210 is used in two board, there should be one chip definition file with all possible GPIO on that *chip* then different board files for the various pins that are available and their special names
e.g. the SAMA5 pin definition https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/sama5/pin.py
and then the GiantBoard, which uses that chip: https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/board/giantboard.py

ladyada:
ok thanks, releasing a new version now - you can pip3 it once updated in pypi and try it!

PR #92, opened 2019-03-18
Include adafruit_blinka.microcontroller.bcm283x.pulseio in package

### LINKED ISSUE #91: adafruit_blinka.microcontroller.bcm283x.pulseio is not packaged
adafruit_blinka.microcontroller.bcm283x.pulseio is missing from the package on PyPI.

`__init__.py` is missing from `adafruit_blinka/microcontroller/bcm283x/pulseio` so `setuptools.find_packages("src")` does not find the package.

Additionally, special handling is needed for the `libgpiod_pulsein` program since it is not python code.

cpforbes:
Fixes #91 by adding `__init__.py` to the package and adding libgpiod_pulsein to setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,9 +30,10 @@ setup(
     url='https://github.com/adafruit/Adafruit_Blinka',
     package_dir={'': 'src'},
     packages=find_packages("src"),
-    # This seems to override find_packages above - I suspect but don't know for sure that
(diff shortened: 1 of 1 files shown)


ladyada:
awesome, will bump! let us know if it still doesnt work after new version's out :)

PR #95, opened 2019-03-22
add I2C and SPI singletons for some drivers

ladyada:
will default to the names we have for I2C and SPI default pins


ladyada:
ok - it was tested on a raspi

ladyada:
please bump a release and you can reply/close the issue

kattni:
Will do!

ladyada:
https://github.com/adafruit/Adafruit_CircuitPython_ServoKit/issues/7#issuecomment-475596473
as well

PR #97, opened 2019-03-25
Allow installation on Windows

MatthewWilkes:
While Blinka does not currently work on Microsoft Windows, it should be possible to install it as managing dependencies with pipenv requires that no install-time errors occur.

The current blocker to installation is a dependency on `sysv_ipc` which only works on POSIX-like operating systems. This change marks the dependency as being conditional on the install target not being Windows (this does not affect people running under cygwin or Windows Subsystem for Linux).

This does not add support for Windows, any attempt to use this code on Windows will result in exceptions being raised or non-functional code. This just makes it possible to manipulate pipenv environments that contain a transitive dependency on Blinka.

For more information on context see https://github.com/adafruit/Adafruit_Python_DHT/pull/113


ladyada:
woohoo! thankx :)

PR #115, opened 2019-05-16
Change PINs to OrangePI R1 spec

kouliss:


ladyada:
hi dont change existing board definition. you can make a new one but don't break things for existing customers!

kouliss:
Sorry, I updated it with new files and put the original ones back

ladyada:
thanks! @makermelissa if you want to order one of  these and check this PR, they seem to be available on amazon

makermelissa:
Ok. Will do.

ladyada:
@kouliss dont forget to add a PR for platform detect as well!

makermelissa:
I should be able to test it this weekend.

kouliss:
The PR for platform detect is done.
@makermelissa thanks for testing, if I can help with something to test, please let me know.

src/adafruit_blinka/board/orangepir1.py
```diff
@@ -0,0 +1,33 @@
...
+PA10 = pin.PA10
+
+PA13 = pin.PA13
+SPI1_CS = pin.PA13
+PA14 = pin.PA14
+SPI1_CLK = pin.PA14
+PA2 = pin.PA2
+UART2_RTS = pin.PA2
+PA18 = pin.PA18
+TWI1-SCK = pin.PA18
```
makermelissa: This needs to be changed to an underscore.

kouliss:
OK, I changed the board.py for loading the correct files, but I am absolutely lost with PIN configuration....

makermelissa:
What were you using as a reference for the pin configuration initially?

makermelissa:
Never mind, I figured out what you most likely used. Probably something like http://dl.sumnic.com/download/OrangePi-R1.pdf. You'll need to add something like:
```
SCLK = pin.PA14
MOSI = pin.PA15
MISO = pin.PA16
```
Looks like there are a bunch more in the document I linked to.

makermelissa:
I'm getting the following error.
```
  File "blinkatest.py", line 16, in <module>
    spi = busio.SPI(board.SCLK, board.MOSI, board.MISO)
  File "/usr/local/lib/python3.5/dist-packages/busio.py", line 90, in __init__
    format((clock, MOSI, MISO), spiPorts))
NotImplementedError: No Hardware SPI on (SCLK, MOSI, MISO)=(14, 15, 16)
Valid SPI ports:((0, 66, 64, 65),)
```
My guess is there isn't something correctly defined in the microprocessor file. I'll need to dig into this a little more to see what needs to get done to finish this up.

makermelissa:
I went ahead and committed the necessary changes directly to your repo. It should be good now and we can fix in another PR if not.

makermelissa:
By the way, the changes I made were that the additional pin file created in the microprocessor directory was unnecessary, so I removed it and I added the second SPI port info to the existing pin file.

PR #116, opened 2019-05-18
Update Giant Board alt pins names and add sysfs pwm support

Groguard:
This pull request:
- Add alternate pin names
- Add sysfs pwm pins
- Update pulseio.py to include Giant Board
- Fix import errors for sysfs pwm


ladyada:
im a bit ambivilent about sysfs_pwmout.py  changes cause id have to go back and make sure they didnt break the Coral - can you change this to not include that change and do that as a different commit?

Groguard:
Yeah, I wasn't sure if I should include that here or not.

PR #129, opened 2019-06-06
Removed Odroid C2 from pulseio

makermelissa:
Since there's nothing in `/sys/class/pwm` and no overlays in the latest mainline kernel release of armbian, I'm going to remove this for now.


ladyada:
agreed!

PR #132, opened 2019-06-11
Added init file to amlogic folder

makermelissa:
I added __init__.py to amlog folder because PyPI was seeing it as an empty folder and not including it, so Odroid C2 wasn't working.


ladyada:
i forget this all the time :(

PR #134, opened 2019-06-18
Use GPIO value-names in bcm283x/pin.py

lurch:
...instead of relying on the fact that Pin.LOW is the same as GPIO.LOW
Also small tidyup to the mode-setting in init

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -28,10 +28,10 @@ class Pin:
     def init(self, mode=IN, pull=None):
         if mode != None:
             if mode == self.IN:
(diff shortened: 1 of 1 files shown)


makermelissa:
Hi, what does this change fix if Pin.LOW and GPIO.LOW are the same? Also, which board or boards did you test this on?

lurch:
It doesn't "fix" anything specifically (i.e. there's no bug in the current code), since as you say Pin.LOW and GPIO.LOW are currently the same. It just feels more "semantically correct" IYSWIM.

Feel free to close if you feel that it isn't worth it.

makermelissa:
Closing this because I don't want to inadvertently break something with this change and there hasn't been a response to @tannewt's great suggestion.

PR #139, opened 2019-07-19
Added Raspberry Pi Compute Module Support

makermelissa:
Tested a couple of the new GPIO pins.


computek45:
Are these changes to python and circuit python libraries?

makermelissa:
I'm not sure what you're asking. The changes affect Blinka and now it shouldn't complain that the board isn't supported and it should actually be able to use the libraries.

PR #142, opened 2019-08-02
Update writebytes to writebytes2 for long buffers

makermelissa:


ladyada:
lets see what breaks :D

PR #147, opened 2019-08-13
Partial work for adding hifive unleashed to Blinka

fede2cr:
Pins from page 25.

https://sifive.cdn.prismic.io/sifive%2Ffa3a584a-a02f-4fda-b758-a2def05f49f9_hifive-unleashed-getting-started-guide-v1p1.pdf

### DIFF
--- a/src/adafruit_blinka/board/orangepipc.py
+++ b/src/adafruit_blinka/board/orangepipc.py
@@ -1,6 +1,6 @@
 """Pin definitions for the Orange Pi PC."""
 
-from adafruit_blinka.microcontroller.allwinner_h3 import pin
+from adafruit_blinka.microcontroller.allwinner.h3 import pin
 
 PA12 = pin.PA12
 SDA = pin.PA12
--- a/src/adafruit_blinka/board/orangepir1.py
+++ b/src/adafruit_blinka/board/orangepir1.py
@@ -1,6 +1,6 @@
 """Pin definitions for the Orange Pi R1."""
 
-from adafruit_blinka.microcontroller.allwinner_h3 import pin
+from adafruit_blinka.microcontroller.allwinner.h3 import pin
 
 PA12 = pin.PA12
 SDA = pin.PA12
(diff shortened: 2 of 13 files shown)


makermelissa:
Hi, I'm not sure if you were already planning to do so, be we normally have separate microcontroller and board files so that if multiple boards use the same microcontroller it's easy to add the new board.

fede2cr:
AFAIK, the microcontroller is not available as a separate device, only included in the HFU processor.
The pins are in microcontoller/ as I was not sure where would be a good place to do it.

makermelissa:
Ok, so the board is really just an abstraction of the microcontroller. Think of the microcontroller as defining the functions that the MCU can do and the board as a way to route which microcontroller pin to which GPIO pin. That way it can be reconfigured very easily.

For instance, let's say we have two boards that use the exact same microcontroller, but the GPIO pins on the header are arranged differently. We can use the same microcontroller file, but have different board files to tell which microcontroller pin goes to which GPIO header pin.

So we define all the pins in microcontroller and then define the board by pointing to the defined pins inside of the microcontroller you defined. Since you already have the pins figured out, you have a lot of the hard stuff done. You can take a look at how I did the dragonboard 410c as an example.

fede2cr:
Does this look ok?

makermelissa:
Almost. Make sure you are also checking for the board/microcontroller in these files:
`src/busio.py`
`src/digitalio.py`
`src/microcontroller/pin.py`
`src/microcontroller/__init__.py`

makermelissa:
Hi, just wanted to check in with you since you're like 98% done.

makermelissa:
Not sure why it auto closed... Was trying to squash new commits, but too many working pieces probably.

makermelissa:
Looks like I have the correct file changes captured on a branch, so I'll just do this from my repo.

makermelissa:
I can't re-open this, so I continued it in #220.

PR #148, opened 2019-08-15
Jetson SPI support and Jetson.GPIO requirement

swarren:
Support SPI controllers on Jetson.
Depend on Jetson.GPIO and stop hard-coding its expected installation path.
Add missing D7 pin definition on Jetson TX2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 Adafruit-PlatformDetect
 Adafruit-PureIO
+Jetson.GPIO; platform_machine=='aarch64'
 RPi.GPIO; platform_machine=='armv7l' or platform_machine=='armv6l'
 rpi_ws281x>=4.0.0; platform_machine=='armv7l' or platform_machine=='armv6l'
 spidev; sys_platform == 'linux'
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
     install_requires=[
         "Adafruit-PlatformDetect",
         "Adafruit-PureIO",
+        "Jetson.GPIO; platform_machine=='aarch64'",
         "RPi.GPIO; platform_machine=='armv7l' or platform_machine=='armv6l'",
         "rpi_ws281x>=4.0.0; platform_machine=='armv7l' or platform_machine=='armv6l'",
         "spidev; sys_platform=='linux'",
--- a/src/adafruit_blinka/board/jetson_nano.py
+++ b/src/adafruit_blinka/board/jetson_nano.py
@@ -29,3 +29,17 @@ D24 = pin.B07
 D25 = pin.B05
 D26 = pin.B04
 D27 = pin.B06
+
+CE1 = D7
+CE0 = D8
+MISO = D9
+MOSI = D10
+SCLK = D11
+SCK = D11
+
+CE1_1 = D23
+CE0_1 = D24
+MISO_1 = D25
+MOSI_1 = D26
+SCLK_1 = D27
+SCK_1 = D27
--- a/src/adafruit_blinka/board/jetson_tx1.py
+++ b/src/adafruit_blinka/board/jetson_tx1.py
@@ -29,3 +29,10 @@ D24 = pin.X00
 D25 = pin.P16
 D26 = pin.X03
 D27 = pin.E06
+
+CE1 = D7
+CE0 = D8
+MISO = D9
+MOSI = D10
+SCLK = D11
+SCK = D11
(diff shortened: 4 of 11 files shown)


swarren:
Note that I haven't tested the change to depend on Jetson.GPIO via setup.py yet; I need to investigate how to test setup.py changes locally in a virtual environment...

swarren:
I've now tested the setup.py change and it works fine.

makermelissa:
Thanks!

ladyada:
@swarren heya - hardware SPI is super nice for displays which would allow us to do projects like 
https://learn.adafruit.com/running-tensorflow-lite-on-the-raspberry-pi-4
on the NANO and others. 

as it is, the SPI enabling instructions are not going to be possible for people to run because the vast majority of folks do not have a secondary Ubuntu machine, and you can't run the updater within a VM (and thats a huge setup task with lots of failure risk)

is there any user-land-y ways to setup the pinmux or send SPI command? for raspberry pi there's a tool you can compile that lets you poke the pinmux registers
https://learn.adafruit.com/adding-basic-audio-ouput-to-raspberry-pi-zero/pi-zero-pwm-audio#changing-the-gpio-alts-2-24
as is, i dont forsee people/students will be able to follow the current instructions to enable SPI and thus wont be able to add display outputs.

swarren:
Yes, we're aware there are a number of complex steps required to configure pins for SPI, I2S, ... We're working on ways to simplify this, but at present, the only supported mechanism is to reflash a new DTB and U-Boot with an updated pinmux table derived from the pinmux spreadsheet. Hopefully we'll have something simpler in place in the future.

ladyada:
ok can y'all provide pre-compiled dtb's so that the cross-compile steps can be skipped

swarren:
The only DTBs we release currently are those that configure all the pins as GPIO inputs. I don't believe we have any plans to release alternative DTBs; hopefully we'll release simpler tools in the time-frame of releasing any updates.

ladyada:
? i thought there were dto for the SPI/I2S? im just tryin' to think of ways/steps that we can simplify it now :)

swarren:
The Jetson SW stack doesn't support DT overlays. (RPi does, but this is all very board-specific SW, and the DT overlay handling logic is not present in Jetson SW.)

ladyada:
sorry, wrong wordin' on my part :) can you provide the precompiled dtb, essentially skipping the x-compile step? if not is it a technical reason or a legal reason

swarren:
Semi-technical, semi-project-management.

With SPI/PWM(x2)/I2S(x2?) to support across 4 boards, there are simply too many combinations to deliver pre-compiled DTBs for each one. We would also need to provide a U-Boot binary for each combination (or at least some; not all boards use U-Boot) since that performs some of the pinmuxing without reference to the DTB.

Besides, to actually make use of the custom DTBs/U-Boots, a separate host PC would still be required to flash them onto the system, so it wouldn't really simplify life much.

ladyada:
how about we start with just the jetson nano, with a DT that has SPI and I2S enabled on the 'classic' Pi pin headers?

swarren:
For L4T feature requests, you'd best contact NVIDIA management; I'm just an engineer. A request on our support forums (https://devtalk.nvidia.com/default/board/139/embedded-systems/1) might be enough, but probably best is if you try the product contacts you have at NVIDIA.

ladyada:
yeah we've gone down that path :) looks like someone contributed 
https://github.com/gtjoseph/jetson-nano-support
still too difficult for most people. it doesn't seem like there's a solution at this time. we'll tell folks that hardware SPI is essentially unsupported for now

swarren:
Just to close the loop. I can confirm that we're actively working on a tool to simplify pinmux configuration on Nano. It won't require the use of a host PC to flash the device, or for any other purpose. I can't confirm (a) which L4T release will include this (it could be the next release, or some other release), (b) when any future L4T release will be, (c) any details about how the tool will work. While that won't help anyone right now, hopefully it gives hope for the future!

makermelissa:
Awesome. When it's available, we can update the guide.

ladyada:
yay that will be great for all the TFT displays we're doing :)

augusttreu:
I just got this result:

hello blinky!
  File "/home/pi/blinkatest.py", line 7, in <module>
    led = digitalio.Digital.InOut(board.D18)
AttributeError: module 'digitalio' has no attribute 'Digital'
>>> 
I ran this on a raspberry pi 2 model B following the instructions in "Circuit Python on Linux and Raspberry Pi" from the Adafruit Learning System".  Can someone direct me to another way to test led on any Pi pin? (New-Kid August, actually a very old man)

jerryneedell:
try  ```led = digitalio.DigitalInOut(board.D18)```  (no "." after Digital)

PR #152, opened 2019-09-06
Orangepizero

hansoostendorp:
This change will add support for the Orange Pi Zero.


makermelissa:
Yep exactly. Thanks for adding this. I'll re-run the test once the updated platform detect is published. After that, I have an Orange Pi Zero on hand and will give it a test.

hansoostendorp:
> Tested successfully. Great job on adding this. :)

Thanks!

PR #158, opened 2019-09-20
Add basic FT232H support.

caternuson:
Adds basic I2C, SPI, and GPIO functionality using FT232H. 

* Relies on [pyftdi](https://pypi.org/project/pyftdi/) library for actual FT232H control.
* To use, set environment variable `BLINKA_FT232H`. Then proceed as normal.
* Pin names match silk on FT232H breakout.
* Currently no support for multiple FT232H's. The first one found on the bus is used, via URL `ftdi:///1`.
* GPIO access is provided by I2C and SPI specific interfaces. To make this transparent, the class level `ft232h_gpio` in `microcontroller.ft232h.pin.Pin` is used to hold the current one. This is then changed in the `I2C` and `SPI` class constructors. Is there a better way?
* Pins D0-D3 are simply ignored for GPIO. D3 could be used in I2C mode, but probably not worth confusion and extra code for one extra GPIO pin.
* The FT232H has dedicated CS pins for SPI, must have at least 1 and can have up to 5. This has been set to 1 but is not actually used. Use any of the GPIO pins as a CS, i.e. the typical CircPy way.
* Helpful pinout table [here](https://github.com/eblot/pyftdi/blob/master/pyftdi/doc/pinout.rst)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ RPi.GPIO; platform_machine=='armv7l' or platform_machine=='armv6l'
 rpi_ws281x>=4.0.0; platform_machine=='armv7l' or platform_machine=='armv6l'
 spidev>=3.4; sys_platform == 'linux'
 sysv_ipc
+pyftdi>=0.30.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ Jetson.GPIO; platform_machine=='aarch64'
 RPi.GPIO; platform_machine=='armv7l' or platform_machine=='armv6l'
 rpi_ws281x>=4.0.0; platform_machine=='armv7l' or platform_machine=='armv6l'
 spidev>=3.4; sys_platform == 'linux'
-sysv_ipc
+sysv_ipc; sys_platform == 'linux'
 pyftdi>=0.30.0
--- a/src/adafruit_blinka/board/ftdi_ft232h.py
+++ b/src/adafruit_blinka/board/ftdi_ft232h.py
@@ -1,0 +1,21 @@
+from adafruit_blinka.microcontroller.ft232h import pin
+
+D4 = pin.D4
+D5 = pin.D5
+D6 = pin.D6
+D7 = pin.D7
+C0 = pin.C0
+C1 = pin.C1
+C2 = pin.C2
+C3 = pin.C3
+C4 = pin.C4
+C5 = pin.C5
+C6 = pin.C6
+C7 = pin.C7
+
+SDA = pin.SDA
+SCL = pin.SCL
+
+SCLK = pin.SCLK
+MOSI = pin.MOSI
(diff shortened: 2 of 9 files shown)


ladyada:
@caternuson please add pyftdi to requirements :)

ladyada:
thanx - dya by chance have a mac to test on?

siddacious:
in addition to the `pyftdi` requirement add, I had to `brew install libusb` when I got this issue
blinka_ftdi

After I installed it, I ran into this:
blinka_not_supported

I verified that when I installed, I got the updated `platform-detect`

caternuson:
Odd. I modified that line to actually print out the `board_id`. So you're getting blank, or maybe `None`? Did you set `BLINKA_FT232H` before running?

The guide will hopefully cover all that setup stuff. Similar to:
https://eblot.github.io/pyftdi/installation.html#prerequisites
Thanks for documenting what you've had to do in your case.

@ladyada Sry, no mac for me. Currently only linux.

siddacious:
I did set the `BLINKA_FT232H` env var. I'll poke some more after I rest a bit

caternuson:
Try running this interactively and see what gets shown on your system:
https://eblot.github.io/pyftdi/installation.html#post-installation-sanity-check

siddacious:
ok @caternuson this was user error. I forgot/didn't notice that your code was on a branch. The only extra hoop I had to jump through was `pip install -e requirements.txt` which was probably just me being unfamiliar with python and blinka.

The only other thing I noticed was trying to set a pull up on C0, I got:`Pin has no attribute PULL_UP` or something similar. Again this may just be a Blinka thing (or a Bryan thing).

I successfully tested SPI and I2C on D0-4 and `digitalio` on D7 and C0. I can test more exhaustively if need be.

It looks like i2c scanning isn't working/supported?
image

siddacious:
Forgot to say: THIS IS SUPER COOOL! Thanks for you work, this is going to be really great

ladyada:
@caternuson can ya add `scan()` - its a useful thing!

caternuson:
Sure, I'll take a look. I think I just missed that.

@siddacious Awesome progress. Thanks for testing.

ladyada:
im testing on windows now - libusb is not fun here either but i have notes

ladyada:
in yoru guide you may want to tell them to run
```
import usb
import usb.util
dev = usb.core.find(idVendor=0x0403, idProduct=0x6014)
print(dev)
```
after installing prereqs - to veirfy that they found the FTDI device on the USB port

alexwhittemore:
Before I go create an issue, I just want to say here, this is rad AF, makes all sorts of things super easy and awesome, I love it, and thank you! I didn't even consider that prototyping stuff could be MORE rapid, but this is just lovely!

ladyada:
yay thanks :)

PR #160, opened 2019-09-26
Chunkify FT232H SPI write

caternuson:
pyftdi has a limit of total size of SPI data that can be written in one operation. This breaks larger data up into multiple chunks and writes.

Sloth love Chunk!


ladyada:
1920px-Baby-Ruth-Split

PR #166, opened 2019-10-19
Fix missing attribute exception catch for 'SPI.chip' on RPi

lichttag:
Previous version did not catch `AttributeError: 'SPI' object has no attribute 'chip'`, which occurs, e.g., on RPi3 running Debian Buster.


ladyada:
@makermelissa im actually ok `pass`ing cause no kernel seems to support this thing they say is supported :)

ladyada:
@makermelissa  plz make anotehr PR that always pass's so this doesnt come up :)

makermelissa:
Ok, will do.

lichttag:
> Ok, will do.

Isn't that exactly what I did?

Issue #159, opened 2019-09-23
have a problem useing PWM on pocketbeagle


fclyyxy1999:
**Board:** pocket beagle
Test:  Test Servo on pocket beagle,which should use PWM
**Code:**
`import time

import board

import pulseio

from adafruit_motor import servo

freqency = (6000 * 1000)  # 6MHz

pwm = pulseio.PWMOut(board.P1_36,duty_cycle=(2 ** 15),frequency=freqency,variable_frequency=True)

my_servo = servo.Servo(pwm)

while True:
    for angle in range(0, 180, 5):
        my_servo.angle = angle
        time.sleep(0.05)
    for angle in range(180, 0, -5):
        my_servo.angle = angle
        time.sleep(0.05)`

**Error: AttributeError: module 'pulseio' has no attribute 'PWMOut'.**

I have seen a similar test on https://github.com/s-light/pocketbeagle_python_tests/blob/master/cp_blinka/blinka_pwm_test.py
And I can use PWM on https://github.com/adafruit/adafruit-beaglebone-io-python successfully.

Is it because there is no pocket beagle PWM support？ I saw #101 but I couldn't find any details.

I'm really confused about these Adafruit libraries.Either python3 or I2C is not supported.Could you please help me solve this problem?

Sincerely thanks!!

pdp7:
Sorry, I missed this when it was assigned.  I'll take a look and update.

pdp7:
I was able to reproduce that error with `cp_blinka/blinka_pwm_test.py` by @s-light:  
```
debian@beaglebone:~/pocketbeagle_python_tests$ sudo python3 cp_blinka/blinka_test.py 
Hello blinka!
Digital IO ok!
I2C ok!
I2C_2 ok!
SPI ok!
SPI_1 ok!
done!
debian@beaglebone:~/pocketbeagle_python_tests$ sudo python3 cp_blinka/blinka_pwm_test.py 
******************************************
test for pwm frequency
Traceback (most recent call last):
  File "cp_blinka/blinka_pwm_test.py", line 14, in <module>
    pwm = pulseio.PWMOut(
AttributeError: module 'pulseio' has no attribute 'PWMOut'
```

pdp7:
@makermelissa @ladyada @sommersoft @Groguard:
I'm looking at the Blink implementation of PWM output for generic linux boards:
[...]

PR #168, opened 2019-10-26
Add PWMOut for PocketBeagle

pdp7:
Add PWMOut for PocketBeagle which has the AM3358 SoC to resolve issue #159 

Unfortunately, [the generic linux PWMOut implementation](https://github.com/adafruit/Adafruit_Blinka/tree/master/src/adafruit_blinka/microcontroller/generic_linux) will not work as [the BeagleBoard.org kernel has a patch](https://github.com/RobertCNelson/bb-kernel/blob/am33x-v4.11/patches/drivers/pwm/0001-pwm-Create-device-class-for-pwm-channels.patch) which changes the naming scheme for PWM channels.

The patch changes the naming scheme from `pwmchipX/pwmY` to `pwmchipX/pwm-X:Y`.  The PocketBeagle has these PWM outputs:
```
/sys/class/pwm/pwmchip0/pwm-0:0/period
/sys/class/pwm/pwmchip0/pwm-0:1/period
/sys/class/pwm/pwmchip2/pwm-2:0/period
/sys/class/pwm/pwmchip2/pwm-2:1/period
/sys/class/pwm/pwmchip4/pwm-4:0/period
/sys/class/pwm/pwmchip4/pwm-4:1/period
```
This naming scheme allows uevents to be created when PWM channels are exported, and thus udev can set the ownership and permissions.  Therefore, the user does not need to be root to manipulate the pwm channel.  Unfortunately, the patch never landed in the mainline Linux kernel, so most systems will have `pwmchipX/pwmY` which is the expectation in [the generic linux PWMOut implementation](https://github.com/adafruit/Adafruit_Blinka/tree/master/src/adafruit_blinka/microcontroller/generic_linux).

For reference, this is the kernel patch that is being applied by @RobertCNelson:
[0001-pwm-Create-device-class-for-pwm-channels.patch](https://github.com/RobertCNelson/bb-kernel/blob/am33x-v4.11/patches/drivers/pwm/0001-pwm-Create-device-class-for-pwm-channels.patch)

### DIFF
--- a/src/adafruit_blinka/board/beaglebone_pocketbeagle.py
+++ b/src/adafruit_blinka/board/beaglebone_pocketbeagle.py
@@ -40,10 +40,10 @@ P1_29 = pin.P1_29    # GPIO3_21 - GPIO_117
 P1_30 = pin.P1_30    # UART0_TXD - GPIO_43
 P1_31 = pin.P1_31    # GPIO3_18 - GPIO_114
 P1_32 = pin.P1_32    # UART0_RXD - GPIO_42
-P1_33 = pin.P1_33    # GPIO3_15 - GPIO_111
+P1_33 = pin.P1_33    # GPIO3_15 - GPIO_111 - EHRPWM0B (ehrpwm.0:1)
 P1_34 = pin.P1_34    # GPIO0_26 - GPIO_26
 P1_35 = pin.P1_35    # GPIO2_24 - GPIO_88
-P1_36 = pin.P1_36    # EHRPWM0A - GPIO_110
+P1_36 = pin.P1_36    # EHRPWM0A - GPIO_110 - EHRPWM0A (ehrpwm.0:0)
 
 
 P2_1 = pin.P2_1      # EHRPWM1A - GPIO_50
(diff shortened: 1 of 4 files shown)


pdp7:
Here is a test script:
```python
import board
import pulseio

pins = [ board.P1_36, board.P1_33, board.P2_1, board.P2_3 ]

for pin in pins:
    print("TEST pin {0}".format(pin))
    pwm = pulseio.PWMOut( pin, duty_cycle=(2 ** 15), frequency=2000, variable_frequency=True)
```
and the output:
```
debian@beaglebone:~$ sudo python3 pwm.py
TEST pin P1_36
TEST pin P1_33
TEST pin P2_1
TEST pin P2_3
```

pdp7:
@makermelissa @ladyada @s-light  does this approach look ok?

s-light:
i did not test it.
i just compared your altered sysfs_pwmout.py file to the original.
```
Adafruit_Blinka (master)$ git diff master:src/adafruit_blinka/microcontroller/generic_linux/sysfs_pwmout.py pdp7/master:src/adafruit_blinka/microcontroller/am335x/sysfs_pwmout.py
```

to me the changes seem all good. 
first i asked myself why you did not overwrite only what changed (as makermelissa had suggested)
but for this it seems the `generic_linux/sysfs_pwmout.py` would need to get some changes too:
possible changes to need only overwriting what is really necessary:
- abstract `self._pin_path.format(self._pwmpin)` as function -
   this way you could override it to use your `self._pin_path.format(self._channel,self._pwmpin)`.
   and all usage of the path would be fine..
- also this would not fix the try/except cases you added..
  but eventually it would make sens/does not hurt to move these to the generic file?

PR #169, opened 2019-10-29
Changed error to more appropriate type

makermelissa:
ValueError makes more sense since the feature is actually implemented.

This fixes https://github.com/adafruit/Adafruit_CircuitPython_WS2801/issues/6 in that it doesn't just quit, but I haven't tested with lights.


jepler:
This makes sense to me.  No testing performed.

makermelissa:
Thanks

PR #170, opened 2019-11-01
Add a lock for thread safety when the I2C class is used in a with statement

hamishmb:
This is a pull request for issue #40 at https://github.com/adafruit/Adafruit_CircuitPython_ADS1x15/issues/40#issuecomment-548527569, which it turns out is best to fix in this project.

I have added a thread.RLock so that only one thread can access the ADS as once, preventing deadlocks. I haven't been able to make the original code crash very easily, but I can confirm that it does, after a similar fix downstream in my project resolved the issue. As it is a race condition, I am unsure how to provide a test program to demonstrate the fix, but existing code seems to work fine with this modification.

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -7,6 +7,8 @@ See `CircuitPython:busio` in CircuitPython for more details.
 * Author(s): cefn
 """
 
+import threading
+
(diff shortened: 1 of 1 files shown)


ladyada:
thanks!

PR #176, opened 2019-11-22
Update pin.py

tb5874:
https://www.jetsonhacks.com/nvidia-jetson-nano-j41-header-pinout/

https://docs.google.com/spreadsheets/d/1LsgtB-vEs6yJQ9URbp_P2XIVghCy400udUMC0MHWm5w/edit#gid=1812645714

SDA0 SCL0 and SDA1 SCL1
order modify
with https://github.com/adafruit/Adafruit_Blinka/pull/175

### DIFF
--- a/src/adafruit_blinka/microcontroller/tegra/t210/pin.py
+++ b/src/adafruit_blinka/microcontroller/tegra/t210/pin.py
@@ -66,10 +66,17 @@ class Pin:
         GPIO.cleanup()
 
 # Cannot be used as GPIO
-SDA = Pin('GEN1_I2C_SDA')
-SCL = Pin('GEN1_I2C_SCL')
-SDA_1 = Pin('GEN2_I2C_SDA')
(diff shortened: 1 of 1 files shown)


makermelissa:
Only after merging these, I noticed this breaks the Jetson TX1 which uses the t210 pin file, but that wasn't updated. If you would still like these changes, please submit a new PR that includes the both the changes to t210/pin.py and both board files updated as a single request. Thanks

tb5874:
> Only after merging these, I noticed this breaks the Jetson TX1 which also uses the t210 pin file, but that wasn't updated. If you would still like these changes, please submit a new PR that includes the both the changes to t210/pin.py and both board files updated as a single request. Thanks

Thank you for check up makermelissa.
However i have only jetson nano, so i can't new PR.
i know their GPIO is same, but i think have tx1 and have no tx1 is very different.
if i have change using both board, i test and submit a new PR.

Thanks again, Have good day :)

PR #182, opened 2019-11-25
Added spidev >= 3.4 in setup.py

makermelissa:
Spidev was specified to be >= 3.4 in requirements,txt, but not setup.py which resulted in people having SPIDEV 3.3 installed. This should fix it.


ladyada:
good catch!

caternuson:
oh sweet. thanks for finding that.

Issue #185, opened 2019-11-26
Support for Beaglebone Green Wireless


crbn60:
I have a SeedStudio BeagleBone Green Wireless I would like to use CircuitPython with.

It is detected correctly but not supported:

```console
>>> detector.chip.id
'AM33XX'
>>> detector.board.id
'BEAGLEBONE_GREEN_WIRELESS'
>>> detector.board.any_beaglebone
True
>>> detector.board.BEAGLEBONE_GREEN_WIRELESS
True
>>> import board
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/board.py", line 107, in <module>
    raise NotImplementedError("Board not supported {}".format(board_id))
NotImplementedError: Board not supported BEAGLEBONE_GREEN_WIRELESS
```

What is required here? Make a copy of the [BeagleBone Black definition](https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/board/beaglebone_black.py) and find anything that is not identical?

I believe the required information can be found on Seed's wiki:

http://wiki.seeedstudio.com/BeagleBone_Green_Wireless/#pin-map

I would like to attempt this myself if I can get a few pointers in the right direction?

makermelissa:
Since they both have the same processor according to https://beagleboard.org/green and https://beagleboard.org/black, you would first need to create a board file inside of Adafruit_Blinka/src/adafruit_blinka/board/ for the green. Then you would need to go through the root level modules and have it load the green where appropriate. It should be pretty straightforward, but feel free to ask if you have any specific questions.

makermelissa:
I just noticed I pointed to green instead of green wireless, but it still applies: https://beagleboard.org/green-wireless

crbn60:
With #187 I am able to make a LED blink!

But I'm not sure that is the correct approach to this, so please comment.

pdp7:
@crbn60 Since the Black and the Green Wireless have the same processor and board headers, I think it would be simpler just to have the BBGW use the same module as the Black:
```
elif board_id == ap_board.BEAGLEBONE_GREEN_WIRELESS:
    from adafruit_blinka.board.beaglebone_black import *
```

makermelissa:
Done!

PR #187, opened 2019-11-30
Duplicate Beaglebone Black for Green Wireless

crbn60:
Implements #185.

As far as I can tell, the pinouts are nearly identical to the Beaglebone Back, so I just duplicated. As I learn more about how hardware works, I will come back and tweak the definition as required.

Please let me know if this is not as easy as I think it is.


crbn60:
I concur with both you and will make the change.

I wasn't sure that was going to work as I'm not sure the board pinouts are actually identical, but I guess I can trip over that in the future as necessary.

ladyada:
yayy! @makermelissa wanna add a downloads page entry some time this week?

makermelissa:
Sure. I think there's a few others like the BBB and Beaglebone Pocket that we can add that just don't have guides yet.

hulkco:
This is my recipe to make de BBGW run Blinka!!!
https://1drv.ms/o/s!Al-YhOt00we6zQbR8-qb3Zggbkqy
Sorry i made this in spanish 
:)

makermelissa:
@hulkco, it's now released and part of Blinka, so you may need to update that guide. Also, if you pip install adafruit-blinka, it should install platformdetect as well as some other packages as dependencies.

PR #190, opened 2019-12-02
Blinka will now work with i2c-3 automatically

makermelissa:
I added i2c-3, which uses SDA and SCL as well. I have updated the I2C code to go with the first one that successfully loads rather than the first that pin-matches and have tested this on i2c-1 and i2c-3 with an Adafruit PiOLED. I also exposed Pins D0 and D1 on the board definition because it seems pointless having i2c-0 defined using D0 and D1 if it can't even be used.

### DIFF
--- a/src/adafruit_blinka/board/raspi_40pin.py
+++ b/src/adafruit_blinka/board/raspi_40pin.py
@@ -2,6 +2,9 @@
 
 from adafruit_blinka.microcontroller.bcm283x import pin
 
+D0 = pin.D0
+D1 = pin.D1
+
 D2 = pin.D2
 SDA = pin.SDA
 D3 = pin.D3
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -139,6 +139,5 @@ uartPorts = (
 )
 
 i2cPorts = (
-    (1, SCL, SDA), (0, D1, D0),   # both pi 1 and pi 2 i2c ports!
+    (3, SCL, SDA), (1, SCL, SDA), (0, D1, D0),   # both pi 1 and pi 2 i2c ports!
 )
-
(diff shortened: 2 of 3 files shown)


makermelissa:
Thanks

declanshanaghy:
How can i enable i2c-3? I tried this in /boot/config.txt. 
```
dtoverlay=i2c-gpio,bus=3,i2c_gpio_delay_us=2,i2c_gpio_sda=27,i2c_gpio_scl=28
```
The device appeared but it's not fully functional.

One thing I notice is what i think is the driver is different in the output of `i2cdetect -l`. It's `3.i2c` as opposed to `bcm2835 I2C adapter`
```
pi@tuya-donor:~ $ i2cdetect -l
i2c-3	i2c       	3.i2c                           	I2C adapter
i2c-1	i2c       	bcm2835 I2C adapter             	I2C adapter
```

The device I had attached didn't show up in `i2cdetect -y 3` and the command is very slow to run. From what i can figure out about that it's because the response from each address is timing out.
```
pi@tuya-donor:~ $ time i2cdetect -y 3
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --                         

real	1m52.812s
user	0m0.001s
sys	1m43.048s
```

makermelissa:
Just for the sake of being able to find it again, this PR was created to address enabling I2C-3 instead of I2C-1 with the original issue at https://forums.adafruit.com/viewtopic.php?f=50&t=159207&p=784164#p784164.

PR #192, opened 2019-12-03
Added BEAGLEBONE_BLACK_INDUSTRIAL to board.py

DonPancoe:
It still imports from adafruit_blinka.board.beaglebone_black because it is otherwise
compatible except for the increased temperature range.


pdp7:
Thanks @DonPancoe !

PR #195, opened 2019-12-07
fix mcp2221 i2c scan

ladyada:
errors throw an exception, a little more elegant than trying to check status


caternuson:
I guess I was lucky and never had this show up. Might it ever throw something other than a Runtime? Like maybe an OSError from something in the HID library?

ladyada:
yeah only would happen during a scan() or if you didnt have the sensor connected

caternuson:
Looks good then if you want the scan to complete. But might be OK to have an exception get thrown if nothing is connected? vs. just silently consuming and continuing? Not sure what the better behavior is.

ladyada:
it'll still throw an exception if you try to connect to a device address that isnt responding, it just wont do that during scanning

PR #196, opened 2019-12-09
Update PyPi requirements for MCP2221

caternuson:
@ladyada Here it is, to the extent I understand how the PyPi plumbing works. @kattni look OK to you?


.travis.yml
```diff
@@ -19,6 +19,7 @@ install:
   - pip install -r requirements.txt
   - pip install --force-reinstall pylint==1.9.2
   - pip install circuitpython-build-tools Sphinx sphinx-rtd-theme
+  - sudo apt-get install libusb-1.0
```
ladyada: put *before* you pip install

caternuson: yah. sry. that was silly of me.

Issue #206, opened 2019-12-18
MCP2221 analogio not being found


caternuson:
```console
$ export BLINKA_MCP2221=1
$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import board
>>> import digitalio
>>> import analogio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'analogio'
>>> 
```

Does it need to be added here?
https://github.com/adafruit/Adafruit_Blinka/blob/cac0d65a5abc39dd7b09360de284012a5e53dc31/setup.py#L47
and maybe other parts of the PyPi deploy?

makermelissa:
Quite possibly. I've never added a new module.

caternuson:
Me either. @kattni Got any pointers? `analogio.py` was added as part of the MCP2221 work. Looks like it's not being brought in when pip installing blinka.

kattni:
@caternuson I have no idea. Closest I've come to working with Blinka is testing that libs `pip` install when completing the Python section of guides.

caternuson:
Welp, there's #207 FWIW.

caternuson:
Yeah. It was that simple. Fixed by #207 and released as 3.3.4.

MCP2221 example:
```python
$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import board
>>> import analogio
>>> knob = analogio.AnalogIn(board.G1)
>>> knob.value
24768
>>> knob.value
43840
>>> knob.value
15360
>>> 
```

PR #207, opened 2019-12-19
Additional PyPi and other things for new analogio module

caternuson:
This is a potential fix for #206. But I'm not a PyPi expert by any means, so not 100% sure.

Also updated a few other items to reference new analogio module.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,7 @@ on hosts running micropython. Working code exists to emulate the CircuitPython p
 
 * **board** - breakout-specific pin identities
 * **microcontroller** - chip-specific pin identities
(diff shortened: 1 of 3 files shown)


makermelissa:
I think it's the double colon causing it to fail.

makermelissa:
Ok, it's failing on NotImplementedError at https://github.com/adafruit/Adafruit_Blinka/blob/cac0d65a5abc39dd7b09360de284012a5e53dc31/src/analogio.py#L15 because the default board is not an MCP2221. You may want to change it to work more like pulseio.

caternuson:
Ugh. Just removing for now. Mainly just want to get the PyPi deploy fixed.

PR #212, opened 2020-01-02
Updated to not install spidev automatically on the mips platform

makermelissa:
If we can get the specific for each board such as detecting the Onion running openwrt, we can add specific requirements in another PR like we did for the RPi-GPIO and Jetson-GPIO.


makermelissa:
Hold off before merging, I want to try a couple other tests first...

makermelissa:
Yeah, this is necessary.

PR #215, opened 2020-01-08
Simple refactor to place all AllWinner boards within the same module

yeyeto2788:
Hey there!

Sorry if this PR comes out of the blue but I'm making this refactor in order to have all `AllWinner` chip defined within the same module for further implementations.

I'm planning to add the `A64` on it after the PR pending on the `Adafruit_Python_PlatformDetect` library is merged.

Regards.

P.D: I know nobody asked about this PR but IMHO seems like more organized as in the `amlogic` one.

### DIFF
--- a/src/adafruit_blinka/board/orangepipc.py
+++ b/src/adafruit_blinka/board/orangepipc.py
@@ -1,6 +1,6 @@
 """Pin definitions for the Orange Pi PC."""
 
-from adafruit_blinka.microcontroller.allwinner_h3 import pin
+from adafruit_blinka.microcontroller.allwinner.h3 import pin
 
 PA12 = pin.PA12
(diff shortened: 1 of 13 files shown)


yeyeto2788:
Hello guys! 

I found this https://github.com/rm-hull/OPi.GPIO/blob/master/orangepi/winplus.py and I just wanted to give it a try.

Let me know if there is anything you want me to change.

Regards.

P.D: Also I see that all almost all Orange Pi board have the same chip (http://linux-sunxi.org/Category:H3_Boards) and pinout so we can use it to add more boards on Blinka, also let me know if you want me add another PR adding them.

makermelissa:
Thanks. This PR totally makes sense. I'll review in more detail today.

PR #224, opened 2020-01-11
Moved repository from Travis to GitHub Actions

dherrada:


makermelissa:
`pylint $( find . -path './adafruit*.py' )
  ([[ ! -d "examples" ]] || pylint --disable=missing-docstring,invalid-name,bad-whitespace $( find . -path "./examples/*.py" ))` won't work on this repo because the structure is completely different. We'll have to come up with a different solution. Also, there's a good chance a lot of things will need to be fixed the first time it's run.

siddacious:
I'm largel  to blame here. I had long day brain and when Dylan asked me if Blinka should be migrated to Actions, I said yes and passed along sommersoft's advice for the bundle 🙄 . My mistake. I think Blinka should be migrated as well but as Melissa said, it's going to need a totally different approach.

@dherrada I'm going to close this ticket for now as I don't consider this as part of the library migration. It will need to be migrated but it's different enough that it will be a separate project.

Issue #226, opened 2020-01-16
Add Orange Pi boards


yeyeto2788:
Hey @makermelissa, I'm creating this issue to keep track of the changes.

Since PR on `adafruit_platformdetect` has been merged I'll be adding them here.

Regards.

P.D: If you want, you can assign this issue to me.

yeyeto2788:
PR pending, waiting for the new version of Adafruit_Python_PlatformDetect to be release in order to launch the build again.

yeyeto2788:
Closing this issue since PR has been merged.

PR #228, opened 2020-01-19
Add Orange Pi boards (One, Lite, PC Plus, Plus 2E)

yeyeto2788:
I'm adding the some boards from Orange Pi which have the same chip (H3) as the one already defined.

Since the pinout is the same as the Orange Pi PC was easy to just import it when one of the boards added into Adafruit_PlatformDetect.

This PR correspond to issue #226 

If any change is needed, just let me know.

### DIFF
--- a/src/board.py
+++ b/src/board.py
@@ -76,6 +76,18 @@ elif board_id == ap_board.ORANGE_PI_R1:
 elif board_id == ap_board.ORANGE_PI_ZERO:
     from adafruit_blinka.board.orangepizero import *
 
+elif board_id == ap_board.ORANGE_PI_ONE:
+    from adafruit_blinka.board.orangepipc import *
+
+elif board_id == ap_board.ORANGE_PI_PC_PLUS:
+    from adafruit_blinka.board.orangepipc import *
+
+elif board_id == ap_board.ORANGE_PI_LITE:
+    from adafruit_blinka.board.orangepipc import *
+
+elif board_id == ap_board.ORANGE_PI_PLUS_2E:
+    from adafruit_blinka.board.orangepipc import *
+
 elif board_id == ap_board.GIANT_BOARD:
     from adafruit_blinka.board.giantboard import *
 


yeyeto2788:
I was thinking also to make a refactor based on vendors in the `board` module.

Something like:

```console
├── board
│   ├── beaglebone_black.py
│   ├── beaglebone_pocketbeagle.py
│   ├── binho_nova.py
│   ├── coral_edge_tpu.py
│   ├── dragonboard_410c.py
│   ├── feather_huzzah.py
│   ├── ftdi_ft232h.py
│   ├── giantboard.py
│   ├── __init__.py
│   ├── jetson_nano.py
│   ├── jetson_tx1.py
│   ├── jetson_tx2.py
│   ├── jetson_xavier.py
│   ├── microchip_mcp2221.py
│   ├── nodemcu.py
│   ├── odroidc2.py
│   ├── odroidn2.py
│   ├── orangepi
│   │   ├── __init__.py
│   │   ├── orangepipc.py
│   │   ├── orangepir1.py
│   │   └── orangepizero.py
│   ├── pyboard.py
│   ├── raspi_1b_rev1.py
│   ├── raspi_1b_rev2.py
│   ├── raspi_40pin.py
│   ├── raspi_cm.py
│   └── tritium-h3.py
```

What do you guys think? This could also be applied for other vendors.

yeyeto2788:
Seems like the build is failing due to the changes not publish to pypi from the Adafruit_Python_PlatformDetect :(

makermelissa:
I see what happened here. It got merged, but not released. I just fixed that. :)

makermelissa:
Regarding the refactor, that definitely sounds like a good plan as we continue to add more and more boards.

yeyeto2788:
Hey @makermelissa!

Yes the layout is the same on all boards except for the One and the Lite. As I could read on sunxi they are rotate 180 degrees but the pins exposed are the same.

I'll do the refactor after finishing the one I did for `adafruit_platformdetect`.

PR #230, opened 2020-01-21
Module refactor

yeyeto2788:
Hey @makermelissa,

This are the changes for the refactor done in Adafruit_Python_PlatformDetect, I might have missed the addition of the Onion Omega (which I'm still looking the way to get it compile libgpiod python bindings).

I'll test the changes tonight but as in Adafruit_Python_PlatformDetect this are changes only on naming and it shouldn't affect any logic.

Also I'll try to make the refactor we talked about putting the boards in blinka by vendor before testing so the testing also covers this.


yeyeto2788:
Take into account that this will fail until the `Adafruit_Python_PlatformDetect` is delivered.

For testing purposes we need to install the libraries from fork `pip install git+https://github.com/yeyeto2788/Adafruit_Python_PlatformDetect.git@module_refactor` and also for Blinka `pip install git+https://github.com/yeyeto2788/Adafruit_Blinka@module_refactor`

Once I get home I'll test the changes on the Orange Pi Plus 2E, I would ask you guys to test it just to make sure I didn't screw things up (I shouldn't).

Regards.

makermelissa:
Thanks. Let's not add the Onion Omega in here yet until we can get something working with the python-libgpiod bindings because we may have to go a totally different path like using debwrt if we can't get something working.

yeyeto2788:
Sure, then let me delete those files then.

yeyeto2788:
@makermelissa Changes done :p

src/board.py
```diff
@@ -127,6 +127,9 @@
 elif board_id == ap_board.PINE64:
     from adafruit_blinka.board.pine64 import *
 
+elif board_id == ap_board.ONION_OMEGA:
```
makermelissa: Can you make sure you're updating the correct branch? I still see Onion Omega stuff here.

yeyeto2788: hahaha my fault, I did not push the changes.

makermelissa:
On the Raspberry Pi, while running the following:
```
import board
import digitalio
import busio

print("Hello blinka!")

# Try to great a Digital input
pin = digitalio.DigitalInOut(board.D4)
print("Digital IO ok!")

# Try to create an I2C device
i2c = busio.I2C(board.SCL, board.SDA)
print("I2C ok!")

# Try to create an SPI device
spi = busio.SPI(board.SCLK, board.MOSI, board.MISO)
print("SPI ok!")

print("done!")
```
I'm getting:
```
Hello blinka!
Digital IO ok!
  File "blinkatest.py", line 12, in <module>
    i2c = busio.I2C(board.SCL, board.SDA)
  File "/home/pi/.local/lib/python3.7/site-packages/busio.py", line 18, in __init__
    self.init(scl, sda, frequency)
  File "/home/pi/.local/lib/python3.7/site-packages/busio.py", line 38, in init
    from microcontroller.pin import i2cPorts
  File "/home/pi/.local/lib/python3.7/site-packages/microcontroller/pin.py", line 9, in <module>
    if chip_id == ap_chip.ESP8266:
AttributeError: module 'adafruit_platformdetect.chip' has no attribute 'ESP8266'
```

yeyeto2788:
Ups, seems like I forgot to edit line on `microcontroller`

Can you please add this on on line 3 in `/home/pi/.local/lib/python3.7/site-packages/microcontroller/pin.py`
```python
import adafruit_platformdetect.chip.contants as ap_chip
```

I'll upload the changes now

yeyeto2788:
Sorry for the inconveniences but I'm facing some issues installing libgpiod python bindings on the Orange Pi Plus 2E (https://pastebin.com/dP0TgLUy) so I haven't been able to test the changes for Blinka since it requires more setup. And I do not have more boards than Orange Pis and the Onion Omega :cry:

makermelissa:
It's ok. That's one of the reasons I'm helping test. I have access to lots of boards.

makermelissa:
Reran on both platforms and it tested successfully on both. I'll work on merging this after the CircuitPython meeting.

yeyeto2788:
Wujuuu! Happy to read that thank you very much.

I'll see what happened with compilation issues on Onion Omega and the other Orange Pi boards.

makermelissa:
Ok, it looks like this is still failing in some places. I have merged and released PlatformDetect changes. Here's the relevant section of the build log:
```
Successfully installed Adafruit-PlatformDetect-2.0.0 Adafruit-PureIO-1.0.4 binho-host-adapter-0.1.4 pyftdi-0.42.2 pyserial-3.4 pyusb-1.0.2 spidev-3.4 sysv-ipc-1.0.1
install.3
1.91s$ pip install --force-reinstall pylint==1.9.2
install.4
5.95s$ pip install circuitpython-build-tools Sphinx sphinx-rtd-theme
1.18s$ pylint src/**/*.py
Using config file /home/travis/build/adafruit/Adafruit_Blinka/.pylintrc
************* Module microcontroller.pin
E:  3, 0: No name 'constants' in module 'adafruit_platformdetect' (no-name-in-module)
************* Module microcontroller
E:  3, 0: No name 'constants' in module 'adafruit_platformdetect' (no-name-in-module)
```

makermelissa:
I had to temporarily point Blinka to an older version so other PRs could pass. Please let me know when you're able to take a look at this again. Thanks

yeyeto2788:
Hey @makermelissa, I had the same problem on local machine and I just did a `pip install -U <module> --force-reinstall` and it worked.

Never the less I my guess is that the problem is coming from `setup.py`. 

I'll take a deeper look at it.

yeyeto2788:
Hey @makermelissa,

I have fixed the Adafruit_Python_PlatformDetect library, the error was coming from the packages it should include on distribution.

Here you have the PR https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/59.

Regards.

makermelissa:
I submitted a PR to your repo that will point to the latest release of PlatformDetect and pull in a few PRs that got merged in yesterday. After that's merged, this should pass.

mruz:
On rpi4 and openSUSE I'm getting module not found error from busio.py[37]: `No module named 'machine'`.
I did `pip3 install machine`, but no avail. I'm not sure if it's related to this change.

makermelissa:
@mruz, you shouldn't need to install a machine library. The problem is that for OpenSUSE, it's not being detected correctly with `detector.board.any_embedded_linux`. This bug is probably in the Platform Detect module, so I be sure that's working correctly first. It could be a result of the recent refactor, but I'm not sure.

mruz:
Right. I see there is no revisions on arm64 Linux https://github.com/adafruit/Adafruit_Blinka/issues/99#issuecomment-500067042, so https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/master/adafruit_platformdetect/board.py#L119 doesn't work for me, but `cat /proc/device-tree/system/linux,revision | xxd` would work.

yeyeto2788:
Hello guys! 

Sorry for the late reply (different time zones), 

In addition to what @makermelissa commented, the issue IMHO is coming from:

```python
    def _pi_rev_code(self):
        """Attempt to find a Raspberry Pi revision code for this board."""
        # 2708 is Pi 1
        # 2709 is Pi 2
        # 2835 is Pi 3 (or greater) on 4.9.x kernel
        # Anything else is not a Pi.
        if self.detector.chip.id != chips.BCM2XXX:
            # Something else, not a Pi.
            return None
        return self.detector.get_cpuinfo_field('Revision')
```

Might it be that the 'proc/cpuinfo` does not have a key/attribute called `Revision` on openSUSE? Or what we are getting from it is not being check?

So far when looking for the `_pi_id` it will compare only with keys in `boards._PI_REV_CODES` and if found return the value of it.

yeyeto2788:
We got to the same conclusion. I guess you should open up an issue on https://github.com/adafruit/Adafruit_Python_PlatformDetect and track changes on it since the issue is comming from that library.

PR #236, opened 2020-01-23
add BeagleBone Green

pdp7:
BeagleBone Green has same pin headers as Black


pdp7:
i did not mean to include https://github.com/adafruit/Adafruit_Blinka/pull/236/commits/24df77e78c7f304d55f3ec6b8e983e42340e648a

PR #255, opened 2020-02-22
Added Beaglebone Black Wireless as a device.

123p10:
Beaglebone black wireless was not supported as a device, but I am using it with i2c on my beaglebone black wireless by simply adding it as a possible device.


pdp7:
@123p10 thank you

@makermelissa this change looks good to me

PR #256, opened 2020-02-24
Add support for NVIDIA Jetson NX

swarren:


swarren:
I guess we need to wait for the platform_detect patch to be merged before the Travis build will pass for this change.

makermelissa:
I just pushed a release for Platform Detect. It should be available in about 5-10 minutes.

PR #268, opened 2020-04-03
Updates to Blinka to use the PureIO SPI library

makermelissa:
This should only be merged after PureIO 1.1.0 is release and so it should fail until then. I'm submitting a PR for PureIO testing purposes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-Adafruit-PlatformDetect
-Adafruit-PureIO
+Adafruit-PlatformDetect>=2.5.0
+Adafruit-PureIO>=1.1.0
 Jetson.GPIO; platform_machine=='aarch64'
 RPi.GPIO; platform_machine=='armv7l' or platform_machine=='armv6l'
 rpi_ws281x>=4.0.0; platform_machine=='armv7l' or platform_machine=='armv6l'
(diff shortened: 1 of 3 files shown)


makermelissa:
Excellent. I want to wait until Monday before merging this in case it does break something.

jepler:
I noticed that your branch still has
```
requirements.txt:spidev>=3.4; sys_platform == 'linux' and platform_machine!='mips'
setup.py:        "spidev>=3.4; sys_platform=='linux' and platform_machine!='mips'",
```
should the spidev dependency be removed now?

makermelissa:
Good catch @jepler! Fixed.

AndySchroder:
I'm looking at the page https://learn.adafruit.com/circuitpython-on-raspberrypi-linux?view=all and it says it was last updated on 2021-11-16 18:34:05 -0500, however, it mentions spidev all over it. Is the spidev part of it totally wrong at this point then?

Also, where can I find performance comparisons between PureIO and spidev for a raspi4? Is there a big performance hit taken by switching to a pure python module vs a c extension, or is all the heavy work done by the kernel? What kind of baud rate can I expect with an MCP3008 with PureIO on a raspi4?

makermelissa:
ioctl talks to the kernel, which does the heavy lifting. This change just made it so people didn't have to install another dependency. The guide should probably be polished a little more though.

PR #274, opened 2020-04-20
ODROID-C4: add GPIO and I2C bus number detection

hhk7734:


makermelissa:
Hi, would it be better to have some of this detection code in a separate file in case you want to add it to other boards such as the N2 so that you don't have as much duplication of code?

hhk7734:
G12A, G12B(N2), and SM1(S905X3) are all similar, so I integrated them into one file. #275 
Review it. If there is a problem, I will fix it. :)

makermelissa:
It appears #275 was in lieu of this PR, so is it ok to close this one? Thanks

hhk7734:
Of course! Tanks @makermelissa .

PR #277, opened 2020-04-24
Add fake pin numbers for raspi to allow selecting explicit i2c bus

### LINKED ISSUE #241: Recent support added for RPi I2C Bus 3 breaks use of Adafruit LCD display device on Bus 1
I have both I2C Bus 1 and I2C Bus 3 enabled on the RPi 3 B+.

The Adafruit I2C LCD display for the RPi is connected to Bus 1 and appears at device address 0x20.

I have other unrelated I2C devices on Bus 1 and Bus 3 (but not at address 0x20 on Bus 3).

I recently brought up a new SD card for the RPi, and installed the current Adafruit software to support the Adafruit LCD, which includes adafruit-circuitpython-charlcd and other packages documented at Adafruit and many subordinate packages.

With the latest software, the LCD no longer operates.  With debugging, I determined the root cause is the Adafruit python code is now trying to open the LCD display on Bus 3 by default, whereas with previous versions of the Adafruit software it correctly opened the LCD display on Bus 1 by default.

I see there have been more than a hundred recent changes to Adafruit_Blinka, and in looking over the list, I am suspicious of this one (but haven't proved it):

https://github.com/adafruit/Adafruit_Blinka/commit/f3a89113466b63c41e62e6bc75eeb40c535c6f75

Is that the culprit?  If so, will that change stand, even though it breaks the Adafruit LCD module?

If the change must stand for other reasons, is there a way (or can it be added) so the application level LCD display code can specify Bus 1 as an override to the new Bus 3, that is breaking our code?

Thank you very much for the Adafruit free software, and your consideration how best to solve this problem for Adafruit LCD display module users.

jsharper:
This will allow explicit selection of a specific i2c bus device.  For example, to reference i2c bus 5 (/dev/i2c-5), use pin references SCL5 and SDA5.

Also, fixes #241 by reverting (part of) previous solution (PR #190) of attempting to auto-choose between i2c-1 and i2c-3 using SCL/SDA pin references.  This PR restores the prior behavior of SCL/SDA (which looks to use i2c-1 or i2c-0 depending on the rpi model) and provides an alternate solution for using i2c-3 (via explicit SCL3/SDA3 pin references).  This should hopefully preserve functionality of all existing code (except for any written in the last few months that was using SCL/SDA to select bus i2c-3) while adding flexibility to specify any particular i2c bus (up to i2c-11) for those who have the need.

I've tested this to talk to 8 SSD1306 OLED displays on i2c-3 through i2c-10 via a TCA9548A multiplexer hanging off of i2c-1 on a raspi zero w.  I also tested examples/pi_busio_i2c.py to talk to an MCP9808 without i2c-mux loaded.

### DIFF
--- a/src/adafruit_blinka/board/raspberrypi/raspi_1b_rev1.py
+++ b/src/adafruit_blinka/board/raspberrypi/raspi_1b_rev1.py
@@ -35,3 +35,31 @@ D22 = pin.D22
 D23 = pin.D23
 D24 = pin.D24
 D25 = pin.D25
+
+# Fake pin numbers to explicitly reference I2C bus devices (/dev/i2c-n) that may
+# or may not be backed by specific physical pins (like virtual i2c buses from i2c-mux)
+# eg: to reference i2c bus 5 (/dev/i2c-5), use SCL5 and SDA5
+SDA0 = pin.SDA0
+SDA1 = pin.SDA1
+SDA2 = pin.SDA2
+SDA3 = pin.SDA3
+SDA4 = pin.SDA4
+SDA5 = pin.SDA5
+SDA6 = pin.SDA6
+SDA7 = pin.SDA7
+SDA8 = pin.SDA8
+SDA9 = pin.SDA9
+SDA10 = pin.SDA10
+SDA11 = pin.SDA11
+SCL0 = pin.SCL0
+SCL1 = pin.SCL1
+SCL2 = pin.SCL2
+SCL3 = pin.SCL3
+SCL4 = pin.SCL4
+SCL5 = pin.SCL5
+SCL6 = pin.SCL6
+SCL7 = pin.SCL7
+SCL8 = pin.SCL8
+SCL9 = pin.SCL9
+SCL10 = pin.SCL10
+SCL11 = pin.SCL11
--- a/src/adafruit_blinka/board/raspberrypi/raspi_1b_rev2.py
+++ b/src/adafruit_blinka/board/raspberrypi/raspi_1b_rev2.py
@@ -35,3 +35,31 @@ D23 = pin.D23
 D24 = pin.D24
 D25 = pin.D25
 D27 = pin.D27
+
+# Fake pin numbers to explicitly reference I2C bus devices (/dev/i2c-n) that may
+# or may not be backed by specific physical pins (like virtual i2c buses from i2c-mux)
+# eg: to reference i2c bus 5 (/dev/i2c-5), use SCL5 and SDA5
+SDA0 = pin.SDA0
+SDA1 = pin.SDA1
+SDA2 = pin.SDA2
+SDA3 = pin.SDA3
+SDA4 = pin.SDA4
+SDA5 = pin.SDA5
+SDA6 = pin.SDA6
+SDA7 = pin.SDA7
+SDA8 = pin.SDA8
+SDA9 = pin.SDA9
+SDA10 = pin.SDA10
+SDA11 = pin.SDA11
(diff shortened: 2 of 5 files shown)


tannewt:
Why does it matter which i2c bus you choose? The idea of the CircuitPython API is that you don't need to know the exact instance. Specifying a specific instance reduces the portability of code.

makermelissa:
Yeah, for linux boards we don't really need the pin numbers. They're not passed to the initializer. They appear to only be a sanity check, but due to the combination of pins that you can do on the Pi, I think the correct fix would probably be to do away with the pins on linux boards rather than create a ton of fake pins.

jsharper:
@tannewt How can you abstract away which bus your code wants?  For instance, I have 8 physical devices on 8 different i2c buses.  My app code needs to be able to specify which exact bus I'm talking to so I can reference a particular physical device.  The 8 buses are not hanging directly off of separate physical pins on the board, so we can't use the simplistic pin-based selection criteria that CircuitPython API is meant to offer for simple microcontroller-based boards.  Keep in mind that Blinka is a shim to allow use of these libs that were designed for one type of system on other types of devices that are more complex and featureful.  This may naturally lead to hacks in the shim abstraction.

@makermelissa I agree that the ton of fake pins is gross and a hack, but it seems like the "right" way to address this would require changing the API to the I2C initializer to allow direct bus selection.  As I mentioned above, I was under the impression that simple microcontroller devices were the "first class citizen" driving the design of these APIs and therefore the blinka shim may have to bring hacks to the table to solve problems that don't exist on simpler boards.  I'd be ok with the I2C initiializer API being updated to align with more complex systems like raspi, but i'd also totally understand if the CircuitPython architects aren't ok with that since it starts violating the original design purpose and concept (as @tannewt's concern illustrates)

makermelissa:
@jsharper you are correct in that when Blinka was originally created, simple microcontroller devices were the "first class citizen", but it has evolved over time and now single board computers are the first class citizen. I think the code could be modified so that it can ignore the pins for SBCs and still make use of them for simpler microcontrollers.

It would likely involve changing the board definitions so we can do something like just defining the pins None and then the code would see that and just use the bus number. I'm adding a new board at the moment, but will likely take a look at I2C again pretty soon.

jsharper:
Thanks for the clarification.  That's great to hear.

I look forward to a future iteration of CP and Blinka that solves this better, and will keep my hacks for use locally until then :)

Thanks again for the convo.

tannewt:
I replied on the other issue but I also wanted to point out one other option.

The option is to create your own I2C-like object for your specific case. All of the CircuitPython drivers are good about taking in the I2C object. It doesn't check that it's a busio.I2C at all so there's no need to change Blinka.

I agree that Blinka is also a first-class citizen but that doesn't mean it can leave CircuitPython behind. Other libraries are welcome to unveil complexity but that isn't Blinka's role.

ladyada:
@tannewt yah i like that idea, it can be a new superclass that takes only an in int for the bus number (or string to device name, like pyserial?) @makermelissa wanna try that?

makermelissa:
Yeah, I was wanting to take a look at I2C next week after finishing the Rock Pi S.

tannewt:
> @tannewt yah i like that idea, it can be a new superclass that takes only an in int for the bus number (or string to device name, like pyserial?) @makermelissa wanna try that?

@makermelissa Please don't do it in Blinka directly because it'll make the library's purpose less clear. It'd be fine to do as a separate library that Blinka can depend on and then subclass.

PR #282, opened 2020-05-03
Add Orange PI 2 support

trainman419:
Add support for the Orange PI 2 using the existing orange pi pc class.

Depends on https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/75


makermelissa:
This appears to be failing due to a bug in Github Actions. I submitted a fix to use a newer version of checkout, but apparently Github Actions refuses to accept that and still wants to use the old workflow. So I'm going to create a new PR so it will use the new file.

makermelissa:
New PR was #285.

trainman419:
Thanks!

Issue #259, opened 2020-03-03
Timed Out Error with PulseIn on Raspberry pi


hyansuper:
This happens on my  Raspberrypi zero(Buster) and Raspberrypi 3(stretch):
```
Python 3.7.3 (default, Dec 20 2019, 18:57:59)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import board, pulseio
>>> p = pulseio.PulseIn(board.D14)
/usr/local/lib/python3.7/dist-packages/adafruit_blinka/microcontroller/bcm283x/pulseio/libgpiod_pulsein: error while loading shared libraries: libgpiod.so.2: cannot open shared object file: No such file or directory
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/adafruit_blinka/microcontroller/bcm283x/pulseio/PulseIn.py", line 64, in __init__
    message = self._wait_receive_msg()
  File "/usr/local/lib/python3.7/dist-packages/adafruit_blinka/microcontroller/bcm283x/pulseio/PulseIn.py", line 80, in _wait_receive_msg
    raise RuntimeError("Timed out waiting for PulseIn message")
RuntimeError: Timed out waiting for PulseIn message
```
I first notice this error when I was using ultrasonic sensor, see https://github.com/adafruit/Adafruit_CircuitPython_HCSR04/pull/16

yeyeto2788:
@hyansuper did you try with other pin? I just noticed this on your error log `libgpiod_pulsein: error while loading shared libraries: libgpiod.so.2: cannot open shared object file: No such file or directory`

hyansuper:
I don't have UART enabled. I also tried many other pins, same error. Do I have to manually install libgpiod? I tried to install libgpiod with [this script](https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/master/libgpiod.sh) but failed. below is the log running the script
```
Installing build requirements - this may take a few minutes!

Hit:1 http://mirrors.tuna.tsinghua.edu.cn/raspbian/raspbian buster InRelease
Hit:2 http://mirrors.tuna.tsinghua.edu.cn/raspberrypi buster InRelease
Reading package lists... Done
N: Ignoring file 'raspi.list.b' in directory '/etc/apt/sources.list.d/' as it has an invalid filename extension
Reading package lists... Done
Building dependency tree
Reading state information... Done
[...]

PR #287, opened 2020-05-07
Updated libgpiod to reduce CPU Usage

### LINKED ISSUE #210: 100% CPU use of libgpiod_pulsein on Raspberry Pi
Trying to access a DHT22 sensor from a Raspberry Pi 4 using `adafruit_dht` and therefore this library, I observe 100% CPU use of the process `libgpiod_pulsein`.

Simple reproducer:
```
#!/usr/bin/env python3

import adafruit_dht
import board
import time

pin=board.D2

dht_device = adafruit_dht.DHT22(pin)

time.sleep(3600)
```

This issue was already mentioned in https://github.com/adafruit/Adafruit_Blinka/issues/140#issuecomment-526713507 but I think it deserves its own bug report.

Attaching strace to the busy libgpiod_pulsein process shows the following messages in a high frequency:
```
ioctl(4, GPIOHANDLE_GET_LINE_VALUES_IOCTL, 0xbea2b1cc) = 0
gettimeofday({tv_sec=1577794452, tv_usec=413127}, NULL) = 0
msgrcv(32768, 0xbea2b37c, 4096, 1, IPC_NOWAIT) = -1 ENOMSG (No message of desired type)
```

Interestingly, the sensor readout itself works well. Also the deprecated https://github.com/adafruit/Adafruit_Python_DHT library does not show this problem.

I'm running the most recent version of raspbian. libgpiod2 is installed in version 1.2-3+rpi1.

makermelissa:
Fixes #210. Also I updated a message in PulseIn so issues like #259 are more obvious about how to fix for the user.


makermelissa:
I forgot to mention I tested the same binary on a Pi 4 and Pi Zero W.

PR #290, opened 2020-05-15
Enable setting no_cs in case a board does support it

makermelissa:
When trying to fix the RA8875, I noticed it wasn't even trying to set no_cs, so this at least tries.

### DIFF
--- a/src/adafruit_blinka/microcontroller/generic_linux/spi.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/spi.py
@@ -55,13 +55,9 @@ class SPI:
 
     # pylint: enable=too-many-arguments,unused-argument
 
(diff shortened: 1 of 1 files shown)


ladyada:
hi im pretty sure this does not work and was causing kernel complaints which is why we removed it. does it work on a pi?

makermelissa:
Yeah, I tested it on the Pi. Maybe using the original spidev driver was causing errors.

ladyada:
and the CS line doesnt toggle?

makermelissa:
No, it's still toggling on the Pi. We can just close this if you think it's not worth it. I was just thinking it might work on other boards, but now that you mention it, it also might cause problems on other boards.

ladyada:
it will cause problems, and i dont like promising it - lets drop it

PR #294, opened 2020-05-22
ODROID-XU4: add support for Hardkernel Odroid XU4

hhk7734:


makermelissa:
I just pushed a PlatformDetect release with the XU4 detection, so we can try running this again in about 10 minutes.

PR #300, opened 2020-05-29
Fixed pylint to check all files and linted

### LINKED ISSUE #291: pylint is not being run on most files
`.github/workflows/build.yml` contains this pylint check:
```
- name: PyLint
      run: |
        pylint src/**/*.py
        ([[ ! -d "examples" ]] || pylint --disable=missing-docstring,invalid-name,bad-whitespace $( find . -path "./examples/*.py" ))
```

The `pylint src/**/*.py` is not actually matching most of the `.py` files, because `**` is not being interpreted by the shell running here as a wildcard match for multiple directories. (In `bash`, for instance, you have to turn on the `globstar` shell option to enable that.)

An alternative is to use `pylint $( find src -name '*.py' )` or something similar.

When this gets fixed, there will be a lot of pylinting to do :smile:, and `missing-docstrings` will have to get turned off, at least for now.

makermelissa:
Fixes #291.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
         black --check --target-version=py35 .
     - name: PyLint
       run: |
-        pylint src/**/*.py
+        pylint $( find src -name '*.py' )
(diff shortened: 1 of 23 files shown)


dhalbert:
Thanks!

Issue #306, opened 2020-06-11
Update Google Coral SPI number


luke-iqt:
The SPI number has changed with the more recent versions of Mendel. There is a mention of it here: https://coral.googlesource.com/coral-cloud/+/63b4aaad128f36f017b107ae6da902369c2e7561
I think it changed when Mendel went to the Day version.
The path for the 2 SPI ports are: /dev/spidev0.0 and /dev/spidev0.1

I am also trying to get the Adafruit LoRa Bonnet working on the Coral. https://learn.adafruit.com/adafruit-radio-bonnets/pinouts

It looks like some of the Pins are not mapped through Blinka on the Coral:
The RST pin maps to Pi_Pin_GPIO, which is pin 22 and maps to GPIO140.... but GPIO 140 is not map through on Blinka
The CS pin maps to Pi_Pin_GPIO7, which is pin 26 and map to ECSPI_SS1... but that isn't mapped in either... and probably busy doing other stuff.

Do you think it would be possible to use the LoRa Bonnet using Blinka? I can look at mapping some of this in.

luke-iqt:
Should be closed with commit 0ab99d44db2e62b79e57ebb751459ac4b371daab

PR #307, opened 2020-06-11
Correct the spiId for the Coral Board

luke-iqt:
The spiId used for the Coral board changed with Day release of Mendel. This helps correct #306


makermelissa:
Thanks.

PR #313, opened 2020-06-28
Binho Nova performance improvement, use new WHR command which consoli…

fig1010:
…dates multiple commands into one transaction


fig1010:
will run black formatter and check in fix.

fig1010:
Please advice on how to fix the remaining pylint errors as those we pre-existing code and no changes were made to those circuitpython function prototype argument list...

makermelissa:
Yes, please go ahead and use that directive.

PR #316, opened 2020-07-09
Chip select in class SPI of busio library 

emmadep93:
Needed if working with multiple devices.

Useful if dev/spidev0.0 is already used from another device

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -143,7 +143,7 @@ class SPI(Lockable):
     for both MicroPython and Linux.
     """
 
-    def __init__(self, clock, chip_select=0, MOSI=None, MISO=None):
+    def __init__(self, clock, chip_select=None, MOSI=None, MISO=None):
         self.deinit()
         if detector.board.ftdi_ft232h:
             from adafruit_blinka.microcontroller.ft232h.spi import SPI as _SPI
@@ -178,7 +178,10 @@ class SPI(Lockable):
                 and MOSI in (portMosi, None)  # Clock is required!
                 and MISO in (portMiso, None)  # But can do with just output
             ):  # Or just input
-                self._spi = _SPI((portId, chip_select))
+                if chip_select is not None and not isinstance(portId, tuple):
+                    self._spi = _SPI((portId, chip_select))
+                else:
+                    self._spi = _SPI(portId)
                 self._pins = (portSck, portMosi, portMiso)
                 break
         else:
--- a/src/busio.py
+++ b/src/busio.py
@@ -143,7 +143,7 @@ class SPI(Lockable):
     for both MicroPython and Linux.
     """
 
-    def __init__(self, clock, MOSI=None, MISO=None):
+    def __init__(self, clock, chip_select=0, MOSI=None, MISO=None):
         self.deinit()
         if detector.board.ftdi_ft232h:
             from adafruit_blinka.microcontroller.ft232h.spi import SPI as _SPI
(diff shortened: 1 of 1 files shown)


tannewt:
Also note that we already have the bus device library for managing individual devices on a spi bus: https://circuitpython.readthedocs.io/projects/busdevice/en/latest/api.html#adafruit-bus-device-spi-device-spi-bus-device

emmadep93:
I’m using Raspberry Pi 3 model B with RS485 CAN HAT that uses `/dev/spidev0.0` and I want also to use MCP3008 (ADC) to sample CAN-L voltages. For this reason I have followed the instruction here: https://learn.adafruit.com/mcp3008-spi-adc/python-circuitpython and installed all packages required. But MCP3008 wants to use `/dev/spidev0.0` even if I changed cs as CE1 of RPi pin as defined in `board` lib. Note that `/dev/spidev0.0` disappeared from spi list because CAN HAT overlayed in `config.txt` as required in [RS485 manual](https://www.waveshare.com/w/upload/2/29/RS485-CAN-HAT-user-manuakl-en.pdf). The only way to assign `/dev/spidev0.1` to MCP3008 is making that changes in SPI class of `busio` lib.

I searched for solutions on the net without results, so I thought it could be useful also for others. But if you think I’m wrong (and have a solution for my case that I can’t see), please feel free to reject my pull request and forgive me for wasting your time (I’m also a noob in all of this, sorry).

Now I will push a new commit that intends to restore the backward compatibility with CircuitPython.

tannewt:
> I searched for solutions on the net without results, so I thought it could be useful also for others. But if you think I’m wrong (and have a solution for my case that I can’t see), please feel free to reject my pull request and forgive me for wasting your time (I’m also a noob in all of this, sorry).

Happy to have a pull request from you! Is your use timing sensitive? We usually rely on BusDevice to control CS but it can be a bit slower than native SPI.

Looks like this is a conflict between CircuitPython's style of driver (bus is separate from bus device) and the RS485 chip using Linux as the driver. @makermelissa would have a better understanding of the pi and how these two may be able to interoperate.

makermelissa:
Using bus device in Blinka, you can really use any GPIO as the chip select.

emmadep93:
```python
import busio
import digitalio
import board
import adafruit_mcp3xxx.mcp3008 as MCP
from adafruit_mcp3xxx.analog_in import AnalogIn
spi = busio.SPI(clock=board.SCK, MISO=board.MISO, MOSI=board.MOSI)
cs = digitalio.DigitalInOut(board.D5)
mcp = MCP.MCP3008(spi, cs)
```

This is the code I'm using as a reference. I know that with bus device library I can select any pin on RPi as chip select and it is not a problem (`DigitalInOut` method). The problem is in the previous line, when I define the SPI with `busio` library. In particular, I can't pass the `portId` parameter to the constructor of SPI class, so by default it is a constant and not a tuple. This means that it selects `/dev/spidev0.0` and not `/dev/spidev0.1` as I want.
This pull request wants to allow users to pass also the `portId` to `busio` in order to prevent these issues.

tannewt:
How about changing https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/generic_linux/spi.py#L19-L22 so that it picks the first available /dev/spidev? Then, you just pass the file path into the PureIO SPI which will be used directly: https://github.com/adafruit/Adafruit_Python_PureIO/blob/master/Adafruit_PureIO/spi.py#L160-L165

This approach should hopefully "Just Work" and not require changing the outer API.

(Is there a spidevice file that isn't associated with a cs pin?)

ladyada:
id be ok with it picking the first available - alll spidevice's are fixed CS pin, so one has to 'go' :)

makermelissa:
I added this functionality in https://github.com/adafruit/Adafruit_Python_Extended_Bus, which is the correct place to make a change like this.

PR #319, opened 2020-07-14
Fix threading issue that stopped Blinka from working on MicroPython

### LINKED ISSUE #318: MicroPython support broken due to import threading in busio
#170 added an `import threading` to busio that isn't available on micropython.

Making the import conditional setting `threading = None` when not available and use of `self._lock` conditional of `if threading:` works to allow the import...

_(but then I find the old feather huzzah esp8266 that i was going to use this on doesn't have enough ram for the Blinka API shims anyways)_

I assume anyone trying to use this on micropython devices with more available ram might care.

makermelissa:
Fixes #318.

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -7,7 +7,10 @@ See `CircuitPython:busio` in CircuitPython for more details.
 * Author(s): cefn
 """
 
-import threading
+try:
+    import threading
+except ImportError:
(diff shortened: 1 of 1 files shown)


makermelissa:
Tested on Raspberry Pi 4 using eInk weather demo.

PR #339, opened 2020-09-04
Updated for STM32->STM32F405 PlatformDetect chip rename

makermelissa:
This will require https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/97 to be merged in and PlatformDetect release 2.15.1 to be performed before it will pass the checks.


jepler (approved):

jepler:
I forced a re-run but it doesn't help as long as Adafruit_Python_PlatformDetect is not yet published.

makermelissa:
Thanks @jepler, I just published and will do a re-run.

Issue #260, opened 2020-03-04
Add support for multiple FT232H


MoTheItGuy:
Hi there,
So I need some help.  We are trying to use 2 (or more) Adafruit FT232H boards (I bought 10 so far for use in a project) each to control several 5m strips of Neopixels as each FT232H only supports 1 SPI output.

The problem is that the blinka library only seems to support 1 board.

When I connect 2, I get this message.
pyftdi.usbtools.UsbToolsError: 2 USB devices match URL 'ftdi:///1'

I have already used the FTDI utility to name the 2 boards differently, but can't seem to specify that anywhere.

I imagine I'm not the only person that's ever wanted to use more than 1 of these awesome little boards.

Any help is greatly appreciated.

Thanks!
Mo

caternuson:
Correct, the current implementation only supports a single FT232H. And you can't really specify which one, it just grabs the first one found. Are you using the latest version of the library? That search URL looks like it's from an older version. It won't fix the multi-board issue, but you should at least be able to use one of the FT232Hs.

I'd suggest changing the name of this issue to "Add support for multiple FT232H".

mpratt14:
@MoTheItGuy Just so you understand, this is much more than a change to the software. The actual problem is that the two devices have the exact same USB vendor and product ID, so the computer sees the two boards as the same device. 

The actual solution is to program the EEPROM on the board to have a different product ID ([using the FT_PROG tool from FTDI](https://www.ftdichip.com/Support/Utilities.htm#FT_PROG)), and then handle identifying it here in the library...

See page 47 of the datasheet:
https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT232H.pdf

rknobel:
Any progress for this issue? Need help with running two or more FT232H boards on the same Linux PC.

mpratt14:
2 things have to be handled (but I think the first one is ok already in this library, but not in pyftdi)

USB IDs:
https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/master/adafruit_platformdetect/chip.py#L72

https://github.com/eblot/pyftdi/search?q=0x6014&unscoped_q=0x6014
(for each file here ^: adding your new product ID to the existing somehow)

the hardware url:
https://github.com/eblot/pyftdi/blob/master/pyftdi/doc/urlscheme.rst

[...]

PR #349, opened 2020-10-06
Enable support for multiple mcp2221

simallaire:
I am working on a project which requires the use of multiple MCP2221 + MLX90640 on a single host. It looked like it was not possible with the current code, so I made this example which shows how it can be accomplished without modifying any code of this library. You can then create different I2C objects independently, in my case MLX90640, and use them normally.

- Redefine existing classes to be able to connect with device path
- Use MCP2221 which is created on import.
- Connect to other MCP2221, if any.
- Print ids (made for mlx90640)

### DIFF
--- a/examples/mcps_busio_i2c.py
+++ b/examples/mcps_busio_i2c.py
@@ -1,0 +1,30 @@
+import time
+import sys
+import board
+import busio
+
+print("hello blinka!")
+
+i2c = busio.I2C(board.SCL, board.SDA)
+
+print("I2C devices found: ", [hex(i) for i in i2c.scan()])
+
+if not 0x18 in i2c.scan():
+    print("Didn't find MCP9808")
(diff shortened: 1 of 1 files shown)


makermelissa:
It looks like you need to run black on your code.

makermelissa:
For PylLint, you can add `# pylint: disable=too-few-public-methods` to those classes.

makermelissa (approved):
Thanks for adding this.

caternuson:
@simallaire If you want, you could also open an issue to request this general support be added to Blinka. I thought there already was one, but not seeing it. There's already an issue for multiple FT232H's:
https://github.com/adafruit/Adafruit_Blinka/issues/260
Just realize it may take some time, as there are lots of details to deal with, similar to all the discussion in the above thread.

PR #352, opened 2020-10-12
add option to skip reset on MCP2221

eteq:
I have an MCP2221 that I could not for the life of me get to work with Blinka until I by-hand skipped the `_reset` stage. On my system (Ubuntu 20.04 on kernel 5.8.10), for whatever reason the reset process takes forever, sometimes >15 seconds. But for many of my use cases I don't really need to bother with the reset process anyway since I'm often just plugging it once so that I know it's already been reset.

So this PR adds a convenient way to skip the reset phase: set the `BLINKA_MCP2221_RESET_DELAY` environment variable to any negative number.  If you prefer not to overload envars this way I could also change it to ust a variable like `BLINKA_MCP2221_SKIP_RESET` or something like that instead.

### DIFF
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -6,7 +6,7 @@ import hid
 
 # Here if you need it
 MCP2221_HID_DELAY = float(os.environ.get("BLINKA_MCP2221_HID_DELAY", 0))
-# Use to set delay between reset and device reopen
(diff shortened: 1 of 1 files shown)


ladyada:
@makermelissa take a look! im ok with it = we did add reset because the MCP2221 could get into a weird state and it was nice to always have it come up 'right' but for advanced users its a risk they can take :D

makermelissa (approved):
Sounds good and code looks fine.

eteq:
Thanks for the fast review and merge!

makermelissa:
No problem. Thanks for submitting this.

PR #360, opened 2020-10-31
Support for Orange Pi Zero Plus2 H5

twa127:


makermelissa (changes requested):
Make sure you are loading you board in board.py. See https://learn.adafruit.com/adding-a-single-board-computer-to-blinka/getting-gpio-working for more information.

makermelissa (approved):

PR #363, opened 2020-11-05
Compatibility with generic linux devices

LukeSkyKoder:
Hello,
I made these changes because i need to use rfid library from pc with i2c-tiniy-usb adapter. I initially based the check on "detector.board.id == 'GENERIC_LINUX_PC'"
But then I preferred to keep better compatibility by checking if the new "i2c_addr" parameter was passed. (so you can use a USB-I2C adapter on the raspberry too)
Regards

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -27,16 +27,16 @@ class I2C(Lockable):
     for both MicroPython and Linux.
     """
 
-    def __init__(self, scl, sda, frequency=400000, i2c_addr=None):
-        self.init(scl, sda, frequency, i2c_addr)
+    def __init__(self, scl=None, sda=None, frequency=400000, i2c_device=None):
+        self.init(scl, sda, frequency, i2c_device)
 
-    def init(self, scl, sda, frequency, i2c_addr):
+    def init(self, scl, sda, frequency, i2c_device):
         """Initialization"""
         self.deinit()
-        if i2c_addr:
+        if i2c_device is not None:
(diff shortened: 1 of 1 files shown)


ladyada:
what is the exact device you're using to access I2C?

makermelissa:
I'm a little unclear about what this Pull Request accomplishes. Could you explain a little better what your goals were with these changes?

LukeSkyKoder:
> what is the exact device you're using to access I2C?

> I'm a little unclear about what this Pull Request accomplishes. Could you explain a little better what your goals were with these changes?

I am using [I2C-Tiny-USB](https://github.com/harbaum/I2C-Tiny-USB) with the digispark board and a Linux PC/Raspberry PI (for I2C 5V devices, instead of using a TTL / 3.3v logic converter).
With an I2C to USB adapter I cannot use the SCL and SDA pins with the busio.I2C class, but I need to pass the position of the I2C adapter as found in the linux subsystem, for example, if the device is in position 5 (/dev /i2c-5), for i2c connection I use:
`i2c = busio.I2C (i2c_device = 5)`
I was unhappy leaving "scl" and "sda" positional argument and with the name of the new parameter in the first commits.

makermelissa:
I would like to leave the function parameters as they were for compatibility with CircuitPython. You might get some value out of https://github.com/adafruit/Adafruit_Python_Extended_Bus which uses the device rather than the SDA and SCL pins.

PR #372, opened 2020-12-04
Add timout loop after exporting the PWM pin.

scottamain:
This gives udev time to respond to the device event. Without this,
the subsequent attempts to modify pin properties will fail when
depending on udev rules to apply new user permissions.

This code is essentially copied from python-periphery, as was most
of this file (both MIT licenced). Although it appears sergeev added
this loop recently, compared to when this code was originally cloned.

### DIFF
--- a/src/adafruit_blinka/microcontroller/generic_linux/sysfs_pwmout.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/sysfs_pwmout.py
@@ -132,7 +132,7 @@ class PWMOut:
                 if e.errno != EACCES or (
                     e.errno == EACCES and i == PWMOut.PWM_STAT_RETRIES - 1
                 ):
-                    raise PWMError(e.errno, "Opening PWM period: " + e.strerror)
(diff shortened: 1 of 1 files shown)


makermelissa (changes requested):
Thanks for adding this. Could you please remove the print statements? I think only a couple of boards use this, but I'd like to keep the experience consistent.

scottamain:
Sorry Melissa, I meant to remove that. Also reformatted with black now.

makermelissa (changes requested):
Thanks. Now I'm seeing the setting 3 times. :)

scottamain:
Hahaha, yeah you're too fast on the draw here and I didn't have time to save my own embarrassments

scottamain:
Okay, looks clear now. Sorry this PR was premature!

makermelissa:
It's all good. Thanks for making the changes. Stylistically, it looks fine. Did you test it with a board?

scottamain:
Yeah, I tested this with the Coral Dev Board. That's where I had the issue.

makermelissa (approved):
Ok cool, thanks.

PR #374, opened 2020-12-06
Add Coral Dev Board Mini support (MT8167)

scottamain:
Also rename the other Coral board's pin module for consistency.

I tested all GPIOs as output with an LED, PWMs with an LED, I2Cs with Adafruit BME280, and SPI also with the BME280. This does not include UART pin support.

For board detection support, see https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/121


scottamain:
PyLint fails without the board detection here: https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/121

makermelissa (approved):
This looks great! Thanks!

PR #377, opened 2020-12-08
Support pin names in RPI style for Coral boards

scottamain:
This should allow code compatibility across boards when using the
"D#" names, as per the Raspberry Pi header.

This also renames several pin names for the Mini so they actually match
the Mini's datasheet pin names. (Previously, I mistakenly used pin names
from the original Dev Board.) These renames are safe at this point
because the Dev Board Mini has not shipped yet.

For example, I verified that both these work on the Mini for pin 40:
`led = digitalio.DigitalInOut(board.D21)`
and
`led = digitalio.DigitalInOut(board.GPIO39)`


makermelissa (approved):
Ok cool. This is a good change and probably one that should be applied to some other boards. :)

scottamain:
Thanks Melissa! I'm still trying out options to display a friendly pinout description from CLI and will share thoughts in a GH issue soon, but it's kind of on my back burner right now.

makermelissa:
No worries. Thank you for your contributions.

PR #379, opened 2020-12-12
expose the neopixel_cleanup method in the main module.

jamesrusso:
Simple change to expose the cleanup method which is bcm283x module in the main module. This method is called atexit of the process, but there are some scenarios where you might want to clean it up before a process exists. For example, you could be changing the length of the string.

### DIFF
--- a/src/neopixel_write.py
+++ b/src/neopixel_write.py
@@ -23,3 +23,8 @@ else:
 def neopixel_write(gpio, buf):
     """Write buf out on the given DigitalInOut."""
     return _neopixel.neopixel_write(gpio, buf)
+
+
+def neopixel_cleanup():
+    """Cleanup and destroy the neopixel resources"""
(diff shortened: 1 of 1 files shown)


tannewt:
Why not do this automatically after the write is complete?

jamesrusso:
I'd think we don't want tear down the DMA stuff after each write?

jamesrusso (commented):
Here we are just exposing the neopixel_cleanup method from the bcm283x concrete implementation. From what I can tell, this is the only implementation.

tannewt:
How about we do it if the length of the write changes and atexit?

jamesrusso:
sure, I think that would work and sounds reasonable.  I can work on that and submit another PR if you wish.

jamesrusso:
See #383 for an alternative solution.

PR #380, opened 2020-12-12
Fix error with physical pull up resistors

yeyeto2788:
Should close https://github.com/adafruit/Adafruit_CircuitPython_DHT/issues/32 and also help on using other boards that RPI with external pull resistors.

If  something else is needed just let me know and I'll be more than happy to help.


ladyada:
hi why not just catch the exception in the DHT library and print out a warning there? we don't want to change the behavior of setting a pull_up - it _should_ raise an exception if using libgpiod without pull up support?

yeyeto2788:
Hey @ladyada,

I tried to catch the exception on DHT library but the pin does not get initialized even being setup as pull-up in the DHT library. 

The generic pin object does not have the option to pass an argument to tell it that it will be using physical resistor. 

Even passing such argument will need to add some similar conditional check againts this argument.

ladyada:
you could catch the exception here
https://github.com/adafruit/Adafruit_CircuitPython_DHT/blob/master/adafruit_dht.py#L163

yeyeto2788:
But then on lines https://github.com/adafruit/Adafruit_Blinka/blob/master/src/adafruit_blinka/microcontroller/generic_linux/libgpiod_pin.py#L52-L63 you'll get the raise exception and it will not be initialized if "external pull-up" is passed, shouldn't it?

I can also make it be `pull=None` and then it will just run fine. The only thing missing is that we are not keeping track if we use external or internal resistors.

Thank you very much for your quick reply BTW.

Want me to make the changes on the DHT library then?

ladyada:
where is the pin initialized with a pullup in DHT library, we only see it assigned on the line 
https://github.com/adafruit/Adafruit_CircuitPython_DHT/blob/master/adafruit_dht.py#L163

yeyeto2788:
Hey @ladyada,

You were totally right 👍🏽 my apologies for creating a not needed PR. I have file the PR on DHT library (https://github.com/adafruit/Adafruit_CircuitPython_DHT/pull/56)

Regards.

P.D: Want me to close this one?

PR #383, opened 2020-12-17
re-initalize the underlying ws2811 module if the lenght of the buffer…

jamesrusso:
If the length of the buffer changes (meaning the length of the LED string has changed), we cleanup the ws2811 structure and allow a new one to be created.

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/neopixel.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/neopixel.py
@@ -16,18 +16,24 @@ LED_STRIP = None  # We manage the color order within the neopixel library
 # a 'static' object that we will use to manage our PWM DMA channel
 # we only support one LED strip per raspi
 _led_strip = None
+_buf_len = None
 
 
 def neopixel_write(gpio, buf):
     """NeoPixel Writing Function"""
     global _led_strip  # we'll have one strip we init if its not at first
+    global _buf_len # we save the length of the buf, and if it changes we will cleanup and re-initalize.
 
-    if _led_strip is None:
+    if _led_strip is None or _buf_len != len(buf):
+        # This is safe to call since it doesn't do anything if _led_strip is None
+        neopixel_cleanup()
+
         # Create a ws2811_t structure from the LED configuration.
         # Note that this structure will be created on the heap so you
         # need to be careful that you delete its memory by calling
         # delete_ws2811_t when it's not needed.
         _led_strip = ws.new_ws2811_t()
+        _buf_len = len(buf)
 
         # Initialize all channels to off
         for channum in range(2):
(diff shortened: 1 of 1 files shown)


makermelissa (changes requested):
Hi @jamesrusso, first of all I think this is a really good approach. I tested this and it works if the buffer length changes as you stated. However, the original problem persists if the buffer length stays the same.

This gave me an idea. As a test, I took a similar approach and saved a reference to the buffer and used as "is not" comparison which seems to fix the issue even if the buffer length stays the same. Anyway, here's the code segment I got to work if you'd like to give it a try and maybe apply it to your PR:
```
# a 'static' object that we will use to manage our PWM DMA channel
# we only support one LED strip per raspi
_led_strip = None
_buf = None

def neopixel_write(gpio, buf):
    """NeoPixel Writing Function"""
    global _led_strip  # we'll have one strip we init if its not at first
    global _buf # we save a reference to the buf, and if it changes we will cleanup and re-initalize.

    if _led_strip is None or buf is not _buf:
        # This is safe to call since it doesn't do anything if _led_strip is None
        neopixel_cleanup()

        # Create a ws2811_t structure from the LED configuration.
        # Note that this structure will be created on the heap so you
        # need to be careful that you delete its memory by calling
        # delete_ws2811_t when it's not needed.
        _led_strip = ws.new_ws2811_t()
        _buf = buf
```

Also, this will need to have black run on it to pass the checks.

jamesrusso:
Interesting. This will tear down and re-configure when you pass in a different buffer reference (which may or may not have a different length). I think it's also a reasonable approach. My specific use case was changing the LED length.

I'm trying to think of a situation where this might not be what the user desires. Obviously, it'll work for the NeoPixel class since it always uses and updates the same buffer within the same instance. I could see a case where a user has multiple buffers (with different, pre-computed values) and then writes them out by calling neopixel_write. In this case, every call to neopixel_write could cause all that cleanup/re-init code to run. It's likely a non-issue as I'm sure not many people are using this low level library and those that do are likely using the same buffer on each write. 

We could always introduce a named parameter to neopixel_write which defaults to False, to prevents any automatic cleanup and only init if _led_strip is None.

So, happy to incorporate this into the PR (and cleanup the black failures) if this is direction everyone wants to go.

makermelissa:
Yeah, I think the intention of the neopixel_write was just to give a Raspberry Pi compatible version of the same built-in CircuitPython function. Since the problem seems to be limited to the Pi, I think this is the best approach. If there are any users who run into an issue with low level code, we can always address it then, but referring users to making use of the library should solve that issue. Thanks.

makermelissa (approved):
Ok, I made the changes myself. If there are any issues, we should address them in a separate PR.

jamesrusso (commented):
+ 1 from me. Thanks for making this change.

jamesrusso:
Thanks! Sorry I was able to get to it.

makermelissa:
No worries. :)

PR #385, opened 2020-12-31
Support for Orange Pi Zero Plus2 H5

twa127:


makermelissa (approved):
Thanks for adding. Looks good.

Issue #391, opened 2021-01-14
MCP2221 I2C Clock Not Changing


caternuson:
Quick check just looking at clock signal during a scan:
```python
$ python3
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import board
>>> i2c = board.I2C()
>>> i2c.try_lock()
True
>>> i2c.scan()
[96]
```
![Screenshot from 2021-01-13 15-56-40](https://user-images.githubusercontent.com/8755041/104527064-ae353c80-55b8-11eb-9abd-dd97dbf174e5.png)

Hmmm, more like 400k. OK, let's grab the MCP, set it again, and try again:
```python
>>> mcp = i2c._i2c._mcp2221
>>> mcp.i2c_configure(100000)
>>> i2c.scan()
[96]
```
![Screenshot from 2021-01-13 15-58-36](https://user-images.githubusercontent.com/8755041/104527136-e0df3500-55b8-11eb-9184-343a69d62cfb.png)

No change.

The current code in `i2c_configure` simply blasts out the setting. It does no checking of current MCP I2C state nor if the setting actually occurred. So...what is the current state?
```python
>>> mcp._i2c_state()
37
```
It's non-zero for some reason. Let's try an i2c cancel first.
```python
>>> mcp._i2c_cancel()
>>> mcp._i2c_state()
0
>>> mcp.i2c_configure(100000)
>>> i2c.scan()
[96]
>>> 
```
![Screenshot from 2021-01-13 15-59-59](https://user-images.githubusercontent.com/8755041/104527332-50edbb00-55b9-11eb-9be0-0303a5b211ec.png)

There we go. Not sure if this is new or has always been there? Been too long since the initial work on this.

Discovered this behavior while sleuthing #262

ladyada:
probably always been there, we've likely been running it at 400khz forever

caternuson:
Maybe it's this?
https://github.com/adafruit/Adafruit_Blinka/blob/084beca285f3f53e13fca734953d83ceff3b4f3c/src/busio.py#L30
Should we change that to 100k?

ladyada:
yah

caternuson:
With #393 , the clock freq is now 100kHz out of the box. So that behavior is notionally fixed. But there still seems to be an issue changing it with `i2c_configure`. Issue name changed to reflect that.

[...]

PR #393, opened 2021-01-14
Change default I2C clock frequency

caternuson:
Minor changed suggested in #391 , but probably not an actual fix for that specific issue. Current default is 400kHz, but 100kHz is generally accepted as the more "standard" clock speed. So this change is simply to bring in line with that.

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -27,7 +27,7 @@ class I2C(Lockable):
     for both MicroPython and Linux.
     """
 
-    def __init__(self, scl, sda, frequency=400000):
+    def __init__(self, scl, sda, frequency=100000):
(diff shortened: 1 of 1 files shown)


ladyada (approved):

caternuson:
@makermelissa Any concern from you of this propagating out and causing subtle issues anywhere on other platforms? It *should* be benign. But since it's changing the base, want to be extra cautious.

makermelissa:
No, I'm good with it. It matches the default in CP, so maybe it will solve some problems.

caternuson:
OK. Thanks. Merging and :crossed_fingers:

PR #400, opened 2021-01-27
Fix/rockpis pins

luxarf:
First of all: Thank for all the good work your do and especially this project. It is a joy to work with!

I was experimenting with I2C my rock pi s and noticed that not all the I2Cs that are available in the hardware (and broken out to the headers) are usable with adafruit-blinka.
Furthermore, I noticed that the pins for I2C0 are declared as belonging to I2C2.

I referenced the [GPIO Pinout Diagram](https://wiki.radxa.com/RockpiS/hardware/gpio#Hardware_V12) provided by radxa.
To check whether everything works I tested my changes on an Rock Pi S in Hardware Revision V12 (I unfortunately do not have any other revision available) and can confirm that the three available I2Cs are working properly.

What else do I need to add to make this PR workable for you?


ladyada:
wow this is the most polite PR - thank you! :)

makermelissa (approved):
Thanks, this looks good to me. I didn't test, but I don't see any breaking changes.

luxarf:
Thanks! That was fast :-)

PR #404, opened 2021-01-28
update minimum version of Adafruit-PlatformDetect

bluesquall:
Needed since 8273fc6, otherwise you will get:

```
AttributeError: module 'adafruit_platformdetect.constants.chips' has no attribute 'DRA74X'
```


makermelissa (approved):
Thanks

makermelissa:
2.27.0 had a bug in it, so I'm just going with the latest.

PR #405, opened 2021-01-29
Added pwmio module

### LINKED ISSUE #402: pulseio.PWMOut -> pwmio.PWMOut
In CP core we're in the process of moving PWMOut from pulseio to pwmio. In 6.x it's in both places and in 7.x it'll only be in pwmio. Blinka should have it in both places as well. We're moving libraries and learn guides over now.

cc @dherrada

makermelissa:
Fixes #402.


tannewt (approved):
Thank you!

PR #409, opened 2021-02-05
adds pin PG11 i2c, spi, serialUART support for nanoPi Duo2

b0hne:
This adds nanoPi Duo2 to usable boards.


makermelissa (approved):
Thanks for adding this.

Issue #410, opened 2021-02-05
Remove I2C 3 from BCM283x pins


caternuson:
The definition here:
https://github.com/adafruit/Adafruit_Blinka/blob/70623d5844a9ed2e0539d2a1093ae0bd07e201df/src/adafruit_blinka/microcontroller/bcm283x/pin.py#L151
was probably added (see #190) as a quick fix to allow using the software `i2c-gpio` dtoverlay, and for a very specific configuration. There is now this library:
https://github.com/adafruit/Adafruit_Python_Extended_Bus
which should be used with any `i2c-gpio`.

Therefore, we should now remove that bus 3 entry from `i2cPorts`.

Here's the thread that made me look:
https://forums.adafruit.com/viewtopic.php?f=8&t=174803
and others have run into it before #241.

caternuson:
Hopefully fixed with #411

PR #411, opened 2021-02-05
Remove I2C bus 3 from BCM283x

caternuson:
For #410.

Removes an entry that allowed use of a very specifically configured `i2c-gpio` port which could then mask the existing default hardware port, for cases where both were trying to be used.


makermelissa (changes requested):
Could you add a comment about using https://github.com/adafruit/Adafruit_Python_Extended_Bus if the user has alternate pin configs? It'll probably save people from opening another 10 issues.

makermelissa (approved):
Thanks!

Issue #389, opened 2021-01-07
Remove/insert ftdi_ft232h device


thijstriemstra:
After initializing an ftdi_ft232h device with some LEDs and then removing the USB device, causes the following error:

```
Traceback (most recent call last):
  File "hardware_ctl.py", line 70, in check_btn
    reading = button.value
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/digitalio.py", line 151, in value
    return self._pin.value() == 1
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/adafruit_blinka/microcontroller/ft232h/pin.py", line 52, in value
    current = Pin.ft232h_gpio.read(with_output=True)
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/pyftdi/i2c.py", line 331, in read
    return self._controller.read_gpio(with_output)
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/pyftdi/i2c.py", line 890, in read_gpio
    data = self._read_raw(self._wide_port)
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/pyftdi/i2c.py", line 992, in _read_raw
    self._ftdi.write_data(cmd)
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/pyftdi/ftdi.py", line 1775, in write_data
    length = self._write(data[offset:offset+write_size])
  File "/home/thijs/.virtualenvs/foo/lib/python3.7/site-packages/pyftdi/ftdi.py", line 2087, in _write
    raise FtdiError('UsbError: %s' % str(ex)) from None
pyftdi.ftdi.FtdiError: UsbError: [Errno 19] No such device (it may have been disconnected)
```

This is fine, but I'd like to re-initialize the ftdi_ft232h once it's re-inserted into the USB port.

I wrote some pyudev code that detects insert/removal of the ftdi_ft232h, so that part's sorted, but my question is: how do I re-initialize ftdi_ft232h so my python program doesn't have to be restarted?

caternuson:
Try recreating your `pin` objects used to control the LED, which should cause the FT232H configuration stuff to be re-run, since it lives in the `pin` constructor:
https://github.com/adafruit/Adafruit_Blinka/blob/084beca285f3f53e13fca734953d83ceff3b4f3c/src/adafruit_blinka/microcontroller/ft232h/pin.py#L15-L25

[...]

PR #417, opened 2021-02-09
FT232H: save i2c instance in Pin

thijstriemstra:
refs #389


tannewt:
@caternuson could this init be done on import rather than first construction?

caternuson:
As I remember it, there's some weirdness because GPIO access is done through either the I2C or SPI controller. So I2C and SPI do some shenanigans as result.
https://github.com/adafruit/Adafruit_Blinka/blob/e0d98bdea5b0323b1ec9568d43c32d57744cde8f/src/adafruit_blinka/microcontroller/ft232h/i2c.py#L17
https://github.com/adafruit/Adafruit_Blinka/blob/e0d98bdea5b0323b1ec9568d43c32d57744cde8f/src/adafruit_blinka/microcontroller/ft232h/spi.py#L23

But maybe there's a different way to do that which could also address #389.

caternuson:
@thijstriemstra Is this code working for you to get you around issue #389 for your use case?

thijstriemstra:
@caternuson yep, I use it like this:

```python
board.C0.i2c.configure("ftdi:///1")
```

it's not pretty but it works. See #389 for more details.

thijstriemstra:
ps. maybe `ft232h_i2c` would be a better name?

caternuson:
My notional thinking/hope was that something like `UsbTools.flush_cache()` could somehow get used automatically in here to recover after being hot plugged. So nothing would have to change in user code. At least that's the hope.

thijstriemstra:
Is this PR acceptable or anything I can change?

caternuson:
The only change would be to run black on `pin.py`. Looks like some trailing white space got introduced and it's causing the CI to fail.

BUT - This PR may not be the best approach. This approach gives you easy access to the underlying FT232H plumbing, but it still requires (as I understand it) that you do something like
```python
board.C0.i2c.configure("ftdi:///1")
```
in your code. I'd be better if we could find a way to do the same thing, but more internally to the library.

thijstriemstra:
Removed the whitespace.

I wouldn't mind placing the code elsewhere but I'm not familiar with overall architecture of this package. Maybe it can be a follow up PR.

thijstriemstra:
Fixed merge conflict with new package structure.

makermelissa:
You may want to pull in the latest changes, which will pin PyLint to an older version and some (maybe all) of the failed checks should go away.

thijstriemstra:
> You may want to pull in the latest changes, which will pin PyLint to an older version and some (maybe all) of the failed checks should go away.

Done.

thijstriemstra:
Anything else I can do here?

makermelissa:
I think we're just waiting on @caternuson to review it again.

caternuson:
@thijstriemstra Are you using SPI in any of your setups?

thijstriemstra:
No.

caternuson:
Looks like maybe you're not even using I2C? Just GPIO. (based on issue thread).

Have you tried accessing the controller via either I2C or SPI? Those classes do store the controller. So instead of:
```python
board.C0.i2c.configure("ftdi:///1")
```
could do this:
```python
# grab an I2C instance in the normal way
i2c = board.I2C()

# to access the I2cController instance and call configure:
i2c._i2c.configure("ftdi:///1")
```
Even if you don't actually use I2C for anything.

Sorry this PR has been open for so long. My general concern is that this change might break other things in subtle ways.  The use of I2C in the Pin class was just a random choice. It could also have been SPI.

thijstriemstra:
> My general concern is that this change might break other things in subtle ways. The use of I2C in the Pin class was just a random choice. It could also have been SPI.

If there's a better way, like one you suggested, I will give it a try. To be continued.. (in #389)

makermelissa:
@caternuson this PR is pretty old. Are we waiting on something to be able to merge or should we just close it until it's ready?

caternuson:
I think this PR is as ready as it's going to be. It's generally OK in terms of CI, etc. But still concerned about the general approach. The original issue in #389 was only using I2C. The FT232H is a bit weird in how GPIO is accessed thru *either* the I2C or SPI controller ([see comment above](https://github.com/adafruit/Adafruit_Blinka/pull/417#issuecomment-776240865)). And only one should be used at any given time (the FT232H can not do SPI and I2C at the same time). So for someone doing something similar, but using SPI, there still may be an issue.

The general issue is dealing with hot swapping. There's more discussion in the #389 issue thread. Unfortunately, none of the other options discussed there seemed to work. :(

@makermelissa I suggest we close both #417 and #389 for now. Dealing with hot swapping USB is tricky.

@thijstriemstra Thanks for this PR and all the discussion. This code modification seems to have worked to fix your specific issue, so simple solution in your case is to continue using it.

These discussions threads will always be here for general reference.

makermelissa:
Ok @caternuson. I'm fine with closing those issues. I'm still a bit unclear about if you want to merge this PR or just close it.

caternuson:
Closing for now. Not sure best general solution for this at this time.

makermelissa:
Oh, gotcha. I didn't notice this PR was #417.

PR #422, opened 2021-02-25
Fix Pylint duplicate code checker

makermelissa:
This is another library that is designed around having duplicate code.


makermelissa:
Hmm, looks like it's failing on "Cyclic import", which could be a genuine issue. I'll need to take a closer look soon as this is backing up the whole train.

makermelissa:
Ok, this was a MicroPython only thing and it appears to just enumerate boards, so I disabled the check.

ladyada (approved):

PR #424, opened 2021-02-28
Fix #423 FT232H Pin returns NotImplementedError

### LINKED ISSUE #423: FT232H: Pin should raise NotImplementedError if pull is not None
Class Pin (in pin.py) of FT232H should raise _NotImplementedError_ if pull is not None to stay compatible. Additionally PULL_NONE, PULL_UP and PULL_DOWN should be defined. 

I'm using a DHT 22 with Adafruit Circuit Python DHT lib. The lib tests whether internal pull ups are supported. The test breaks without the mentioned changes.

thomas6g:
#423 
Class Pin of FT232H returns NotImplementedError if pull is not None, now.
Constants PULL_NONE, PULL_UP and PULL_DOWN are defined, now.

### DIFF
--- a/src/adafruit_blinka/microcontroller/ft232h/pin.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/pin.py
@@ -8,6 +8,9 @@ class Pin:
     OUT = 1
     LOW = 0
     HIGH = 1
+    PULL_NONE = 0
+    PULL_UP =  1
+    PULL_DOWN = 2
 
     ft232h_gpio = None
 
@@ -36,7 +39,7 @@ class Pin:
             raise RuntimeError("Can not init a None type pin.")
         # FT232H does't have configurable internal pulls?
         if pull:
-            raise ValueError("Internal pull up/down not currently supported.")
+            raise NotImplementedError("Internal pull up/down not currently supported.")
         pin_mask = Pin.ft232h_gpio.pins | 1 << self.id
         current = Pin.ft232h_gpio.direction
         if mode == self.OUT:
--- a/src/adafruit_blinka/microcontroller/ft232h/pin.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/pin.py
@@ -9,7 +9,7 @@ class Pin:
     LOW = 0
     HIGH = 1
     PULL_NONE = 0
-    PULL_UP =  1
+    PULL_UP = 1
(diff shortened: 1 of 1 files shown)


src/adafruit_blinka/microcontroller/ft232h/pin.py
```diff
@@ -8,6 +8,9 @@ class Pin:
     OUT = 1
     LOW = 0
     HIGH = 1
+    PULL_NONE = 0
+    PULL_UP = 1
```
ladyada, line 12: is this UP and DOWN needed? we dont ever use?

ladyada (changes requested):

src/adafruit_blinka/microcontroller/ft232h/pin.py
thomas6g: Hmm, I thought the definitions are needed for the test whether pull-up/down is supported. I'll check tomorrow Morning.

thomas6g: Yes, I think the PULL_UP and PULL_DOWN are needed because module digitalio uses them. So, even if they are not used because pull up/down is not supported by the HW, the are needed for the test of the HW capabilities. 

Please see the digitalio code snippet of class **DigitalInOut** below
```Python
 @pull.setter
    def pull(self, pul):
        if self.direction is Direction.INPUT:
            self.__pull = pul
            if pul is Pull.UP:
                self._pin.init(mode=Pin.IN, pull=Pin.PULL_UP)
            elif pul is Pull.DOWN:
                if hasattr(Pin, "PULL_DOWN"):
                    self._pin.init(mode=Pin.IN, pull=Pin.PULL_DOWN)
                else:
                    raise NotImplementedError(
                        "{} unsupported on {}".format(Pull.DOWN, board_id)
                    )
```

ladyada: ok!

ladyada (approved):

src/adafruit_blinka/microcontroller/ft232h/pin.py
thomas6g: Hmm, on long term it might be an idea to refactor the redundant definitions within the Pin classes. However for the moment the definition should fix the problem.

PR #427, opened 2021-03-05
add microcontroller.delay_us()

dhalbert:
I wrote a [CircuitPython library](https://github.com/dhalbert/CircuitPython_LCD) a long time ago that uses `microcontroller.delay_us()`. Someone wanted to use it on an RPi, and could not use it without modification because `microcontroller.delay_us()` does not exist in Blinka. This adds it, just calling `time.sleep()` to simulate it.

I can't use `time.sleep()` in the original library because it is only accurate to the nearest millisecond. Values smaller than a millisecond are rounded, so, for instance a few hundred microseconds is no delay at all.

### DIFF
--- a/src/microcontroller/__init__.py
+++ b/src/microcontroller/__init__.py
@@ -1,8 +1,15 @@
 """Microcontroller pins"""
 
+import time
+
 from adafruit_platformdetect.constants import chips as ap_chip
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Looks good to me.

PR #436, opened 2021-03-25
Add support for ftdi URLs in the BLINKA_FT232H variable.

renpytom:
This allows the an FTDI URL, like ftdi://ftdi:232h:W00001/1
to be given in the BLINKA_FT232H. If it's used, then that
board is the one that is opened. If the value of the variable
is not an FTDI url, then the default is used.

This makes it possible for multiple processes on the same
computer to talk to different FTDI boards. While talking
to multiple boards from a single process might be possible,
that's outside of the scope of this commit.

### DIFF
--- a/src/adafruit_blinka/microcontroller/ft232h/url.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/url.py
@@ -2,10 +2,11 @@
 
 import os
 
+
 def get_ftdi_url():
     """
     Return the FTDI url to use. If BLINKA_FT232H starts with ftdi:, returns
-    that. Otherwise, returns a default value.  
+    that. Otherwise, returns a default value.
     """
 
     url = os.environ.get("BLINKA_FT232H", "1")
--- a/src/adafruit_blinka/microcontroller/ft232h/url.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/url.py
@@ -13,5 +13,5 @@ def get_ftdi_url():
 
     if url.startswith("ftdi:"):
         return url
-    else:
-        return "ftdi://ftdi:ft232h/1"
+
+    return "ftdi://ftdi:ft232h/1"
--- a/src/adafruit_blinka/microcontroller/ft232h/url.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/url.py
@@ -1,0 +1,16 @@
+"""Support for getting the URL from the BLINKA_FT232H variable."""
+
+import os
+
+def get_ftdi_url():
+    """
+    Return the FTDI url to use. If BLINKA_FT232H starts with ftdi:, returns
+    that. Otherwise, returns a default value.  
+    """
+
+    url = os.environ.get("BLINKA_FT232H", "1")
+
+    if url.startswith("ftdi:"):
+        return url
+    else:
+        return "ftdi://ftdi:ft232h/1"
--- a/src/adafruit_blinka/microcontroller/ft232h/i2c.py
+++ b/src/adafruit_blinka/microcontroller/ft232h/i2c.py
@@ -1,6 +1,6 @@
 """I2C Class for FT232H"""
 from adafruit_blinka.microcontroller.ft232h.pin import Pin
-
+from adafruit_blinka.microcontroller.ft232h.url import get_ftdi_url
 
 class I2C:
     """Custom I2C Class for FT232H"""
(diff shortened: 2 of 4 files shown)


ladyada:
seems nifty to me

@caternuson or @makermelissa one of ya wanna test to make sure it still works?

caternuson:
Tested with a simple I2C scan and works. Could we do this somehow without needing a new file / function? The logic is pretty simple and looks like it could be done with a ternary. And just update i2c/spi/pin.py to use that.

renpytom:
I think it's just a little too complicated for a ternary, as one would have to first read the url into a variable, and then make a decision based on that variable.

If you think the additional file is unnecessary, then maybe it would make sense to put the get_ftdi_url function into adafruit_blinka/microcontroller/ft232h/__init__.py ?

makermelissa:
I see what @caternuson is asking. You could do something like:
```
self._spi.configure(url if url.startswith("ftdi:") else "ftdi://ftdi:ft232h/1")
```

fharding1:
My two cents is that it makes sense to include another file for this.

There should be only one place where the constant `"BLINKA_FT232H"` exists. It would not make sense to define a `get_ftdi_url()` function in the i2c, spi, or pin file and use it from the others. So it belongs in it's own file.

I guess you could shorten the function to:

```python
def get_ftdi_url():
    """
    Return the FTDI url to use. If BLINKA_FT232H starts with ftdi:, returns
    that. Otherwise, returns a default value.
    """

    url = os.environ.get("BLINKA_FT232H", "1")
    return url if url.startswith("ftdi:") else "ftdi://ftdi:ft232h/1"
```

But this is a pretty insignificant change that's not worth much discussion in my opinion.

caternuson:
> There should be only one place where the constant "BLINKA_FT232H" exists.

There is. It only exists as an environment variable.

> It would not make sense to define a get_ftdi_url() function in the i2c, spi, or pin file and use it from the others.

The idea with the ternary is it gets rid of the need for that function and just does the equivalent logic in place. And therefore, gets rid of the need for the extra file.

caternuson:
If the concern is having references to the envvar `BLINKA_FT232H` in more than one place in the library code, then we could at least move the `get_ftdi_url` function into `__init__.py`. 

And then do a relative import in `i2c.py` and `spi.py`:
```python
 from . import get_ftdi_url
 ```

**Comment by fharding1** [comment]
> There is. It only exists as an environment variable.
> If the concern is having references to the envvar BLINKA_FT232H in more than one place in the library code

Yes, that was the concern expressed here: "there should be only one place where the constant `"BLINKA_FT232H"` exists"

> then we could at least move the get_ftdi_url function

Sure, I guess it could go in `__init__.py`.Not exactly where one would think to look, especially since most of them are empty in this project and they aren't really meant to have helper functions.

File: src/adafruit_blinka/microcontroller/ft232h/url.py [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+    Return the FTDI url to use. If BLINKA_FT232H starts with ftdi:, returns
+    that. Otherwise, returns a default value.
+    """
+
+    url = os.environ.get("BLINKA_FT232H", "1")
+
+    if url.startswith("ftdi:"):
+        return url
+
+    return "ftdi://ftdi:ft232h/1"
```
caternuson, line 17: Suggest moving this into `__init__.py` and then using a relative import in the other 3 .py files.

caternuson (changes requested):

caternuson:
OK, want to get this done. Tried pushing those changes, but not allowed. Let's just merge this as is so other PRs can proceed. Can refactor later.

### Test Setup
ft232h_test

### Test program
```python
import time
import board
import digitalio

led = digitalio.DigitalInOut(board.C0)
led.direction = digitalio.Direction.OUTPUT

while True:
    led.value = not led.value
    time.sleep(1)
```

### Single Board
With a single FT232H attached, previous usage still works:
```
(blinka) $ export BLINKA_FT232H=1
(blinka) $ python3 blink.py
```
Which ever board is currently plugged in blinks.

### Multiple Boards
Plug in both FT232H's and then get URLs:
```python
>>> from pyftdi.ftdi import Ftdi
>>> Ftdi().open_from_url('ftdi:///?')
Available interfaces:
  ftdi://ftdi:232h:1/1          (￿￿￿￿￿￿)
  ftdi://ftdi:232h:FT2QJUM7/1   (USB <-> Serial Converter)
```

Can then specify URL to select FT232H to use:
```
(blinka) $ export BLINKA_FT232H=ftdi://ftdi:232h:1/1
(blinka) $ python3 blink.py
```
One board blinks.
```
(blinka) $ export BLINKA_FT232H=ftdi://ftdi:232h:FT2QJUM7/1
(blinka) $ python3 blink.py
```
Other board blinks.

PR #437, opened 2021-03-29
FT2232H support and FT232H/MPSSE refactor

### LINKED ISSUE #308: FT2232H support
Hi! I'm working on a board called [Tigard](https://github.com/tigard-tools/tigard) at the moment, which is based off the FTDI FT2232H. The FT2232H is similar to the FT232H, but with dual MPSSEs. It would be awesome if I could use Tigard (or some of the other FT2232H-based boards such as [timep](https://github.com/Matir/timep)) with CircuitPython/Blinka, and I see there is already support for the FT232H! I'd be happy to take this work on myself, but wanted to ask some questions before throwing a PR at you so my work can hopefully be merged upstream if this is something Blinka wants to support:

There are a few options for implementing AFAICT:

* Make a generic class that's used for both the FT232H and the FT2232H classes (similar to how the "amlogic" folder is structured) . Since the two chips are really similar, I think this makes the most sense. I'm not sure about the name, maybe something along the lines of `ftdi`, `ftdi-mpsse`, or `ftxxxx`?
* Just copy the FT232H class and make a FT2232H class. This is easy and keeps the FT232H code as-is but super redundant.
* Simply modify the FT232H class to look for `ftdi:///1` instead of `ftdi://ftdi:ft232h/1`. This technically works (I have tested it), but you can only use the first MPSSE so it's really not ideal.

I noticed there's also issue #260, which is pretty similar to this, but I don't think makes sense given how Blinka works? The difference is that this issue describes adding support for a new "microcontroller" (the FT2232H) that happens to have two MPSSEs, instead of adding support for using two of the same "microcontroller" (two FT232Hs) each with one MPSSE. However, if support for the latter was ever going to be added, I think a more generic FTDI class would help with making that easier.

I think the first is the clear choice, but I'm not super familiar with this library, so let me know what you think, thanks!

Edit: implemented the first, still needs some more testing but check it out here: https://github.com/tigard-tools/Adafruit_Blinka

fharding1:
This pull request adds support for the FT2232H without breaking any existing functionality with the FT232H, as outlined in #308. So far all the tests I've thrown at my Tigard FT2232H dev board (MCP9808 I2C temperature sensor, GPIO, etc) have worked out great, but if someone wants to verify with some other (SPI) devices (possibly on another OS besides Linux, too) that would be great.

https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/149 accompanies this pull request.

Closes #308


fharding1:
I'm not sure why CI is broken, this code has been formatted with black.

```
➜  Adafruit_Blinka git:(master) python -m black --check --target-version=py35 . All done! ✨ 🍰 ✨
213 files would be left unchanged.
```

fharding1:
CI will be failing without https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/149

k0d:
I literally just started to fix the support for this, and then saw you'd started to do it. Awesome!!

I'm testing it on MacOS, and it doesn't detect my board as it looks for "ftdi://ftdi:ft2232h/1", but mine shows up as 
```
Available interfaces:
  ftdi://ftdi:4232:14:8/1   (Quad RS232-HS)
  ftdi://ftdi:4232:14:8/2   (Quad RS232-HS)
  ftdi://ftdi:4232:14:8/3   (Quad RS232-HS)
  ftdi://ftdi:4232:14:8/4   (Quad RS232-HS)
```
I'm not sure how the naming works, I'm using the dev board (mini-module) which has a ft2232hl I believe, and you're using the *q on your board it seems. So I thought they'd be the same?

k0d:
Ok, so the issue was the EEPROM wasn't being read, but now I've got that working it's still different. I guess I need to change the EEPROM contents. Which makes me wonder how we should deal with different boards with different settings.

fharding1:
> Ok, so the issue was the EEPROM wasn't being read, but now I've got that working it's still different. I guess I need to change the EEPROM contents. Which makes me wonder how we should deal with different boards with different settings.

I was thinking we should probably wait for https://github.com/adafruit/Adafruit_Blinka/pull/436 to be merged first, and then do the same thing for the FT2232H.

k0d:
Sounds like a good plan. That will certainly solve the issue...

makermelissa:
I merged in the PlatformDetect portion, so that should help with getting this fixed up.

caternuson:
Based on this behavior:
https://github.com/adafruit/Adafruit_Blinka/issues/308#issuecomment-665817870
Could this be done by taking the approach of support N FT232H's vs. specifically supporting just the FT2232H? (which would just be the N=2 case)

fharding1:
> Based on this behavior:
> [#308 (comment)](https://github.com/adafruit/Adafruit_Blinka/issues/308#issuecomment-665817870)
> Could this be done by taking the approach of supporting N FT232H's vs. specifically supporting just the FT2232H? (which would just be the N=2 case)

I'm not sure why this is being rehashed in the pull request. Is there something you wanted to add to that discussion?

caternuson:
OK, support will be for FT232H and FT2232H. No support (at this time) for N MPSSE's.

There's a general conflict between this PR and #436 since both are modifying the way the URL gets specified and used. Suggest we get #436 merged first and then refactor this one based on that.

fharding1:
> There's a general conflict between this PR and #436 since both are modifying the way the URL gets specified and used. Suggest we get #436 merged first and then refactor this one based on that.

Yes https://github.com/adafruit/Adafruit_Blinka/pull/437#issuecomment-809704718

fharding1:
Okay, that should do it. A user should set `BLINKA_FT2232H_0`, `BLINKA_FT2232H_1`, ..., `BLINKA_FT2232H_n` for any n interfaces they want to enable custom URLs for. For example, `export BLINKA_FT2232H_1=ftdi://ftdi:ft2232h/2` means that the URL used for `busio.I2C(board.SCL1, board.SDA1)` will be `ftdi://ftdi:ft2232h/2`.

caternuson:
Thanks. Looks good. Sorry this took so long to finally get reviewed, tested, and merged.

Tested that it did not affect current usage with FT232H:
```python
(blinka) $ export BLINKA_FT232H=1
(blinka) $ python3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
>>> import board
>>> import digitalio
>>> led = digitalio.DigitalInOut(board.C0)
>>> led.direction = digitalio.Direction.OUTPUT
>>> led.value = True
>>> led.value
True
>>> led.value = not led.value
>>> led.value = not led.value
>>> 
```
Board shows up and LED blinks as expected.

Only have one FT2232H available for testing, so was not able to test the multiple URL feature. But enabling for basic use with a single FT2232H worked as expected.
```python
(blinka) methuzla@bench:~$ export BLINKA_FT2232H=1
(blinka) methuzla@bench:~$ python3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
>>> import board
>>> dir(board)
['AC0', 'AC1', 'AC2', 'AC3', 'AC4', 'AC5', 'AC6', 'AC7', 'AD4', 'AD5', 'AD6', 'AD7', 'BC0', 'BC1', 'BC2', 'BC3', 'BC4', 'BC5', 'BC6', 'BC7', 'BD4', 'BD5', 'BD6', 'BD7', 'MISO0', 'MISO1', 'MOSI0', 'MOSI1', 'SCK0', 'SCK1', 'SCL0', 'SCL1', 'SCLK0', 'SCLK1', 'SDA0', 'SDA1', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'ap_board', 'board_id', 'detector', 'pin', 'sys']
>>> import digitalio
>>> led = digitalio.DigitalInOut(board.AC0)
>>> led.direction = digitalio.Direction.OUTPUT
>>> led.value = True
>>> led.value = False
>>> led.value
False
>>> led.value = not led.value
>>> led.value = not led.value
>>> 
```
Board showed up, LED blinked, etc.

fharding1:
Thanks!

PR #438, opened 2021-04-03
Added internal I2C bus for the CM4

mebs:
Added I2C bus 10 definition for the Raspberry Pi CM4.
I tested the change on my own CM4 on a custom carrier board and a SSD1306 display and it works fine.

This PR was inspired by this one:
https://github.com/adafruit/Adafruit_Blinka/pull/83

The pins for the I2C bus on the CM4 are described in its datasheet, but the subject is also discussed here:
https://www.raspberrypi.org/forums/viewtopic.php?t=296162

This is my first contribution to an open source project, apologies if I omitted something!

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -153,5 +153,5 @@ uartPorts = ((1, TXD, RXD),)
 i2cPorts = (
     (1, SCL, SDA),
     (0, D1, D0),  # both pi 1 and pi 2 i2c ports!
-    (10, D45, D44), # internal i2c bus for the CM4
+    (10, D45, D44),  # internal i2c bus for the CM4
 )
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -153,4 +153,5 @@ uartPorts = ((1, TXD, RXD),)
 i2cPorts = (
     (1, SCL, SDA),
     (0, D1, D0),  # both pi 1 and pi 2 i2c ports!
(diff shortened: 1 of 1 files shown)


makermelissa:
Hi, with the raspberry pi boards, there seems to be an enormous number of possible permutations with regards to the I2C ports, which is why we wrote the https://github.com/adafruit/Adafruit_Python_Extended_Bus library. It allows you to directly pass the i2c device number instead of defining some pins just so that it can figure out the device number.

mebs:
> Hi, with the raspberry pi boards, there seems to be an enormous number of possible permutations with regards to the I2C ports, which is why we wrote the https://github.com/adafruit/Adafruit_Python_Extended_Bus library. It allows you to directly pass the i2c device number instead of defining some pins just so that it can figure out the device number.

Thanks for the info, this is good to know! I'm surprised it didn't show in my (admittedly quick) research.

However, the CM4 does not have IIC buses on the same pins as the regular Pi models. 
This is not a random, additional hardware IIC configuration, this is the bare minimum to be able to use the protocol on the CM4. 
I think there is value in adding it here (and perhaps adding the right SCL and SDA definitions in `board`) as all CM4 users would systematically have to go through the extended library in the current state of things.

Feel free to close this PR if you disagree!

makermelissa (approved):
> However, the CM4 does not have IIC buses on the same pins as the regular Pi models.
> This is not a random, additional hardware IIC configuration, this is the bare minimum to be able to use the protocol on the CM4.
> I think there is value in adding it here (and perhaps adding the right SCL and SDA definitions in `board`) as all CM4 users would systematically have to go through the extended library in the current state of things.

I see. I just looked at some info online. I didn't realize this was a hardware I2C port, but this PR makes more sense now.

Issue #439, opened 2021-04-05
Support for the original Beaglebone


Cellgalvano:
Hi,

I created a project using Blinka for the PocketBeagle and everything was working fine out of the box! Now I'd like to port it over to a few older original white Beaglebones but the board is not supported.

As far as I understand the SoC definition of the original Beaglebone is (mostly?) equal to the Beaglebone Black and therefore one could simply add another "else if" in boards.py for the board id "BEAGLEBONE" to support the older boards?

Thank you!

ladyada:
seems like it should work - we don't have an older 'Bone on hand, so if y'ad like to add support please send in a PR :)

pdp7:
@Cellgalvano it should work okay, just need to add BEAGLEBONE to:
https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/59edca95e6324678cf8ebef9f0356a30c7943369/adafruit_platformdetect/constants/boards.py#L291

The EEPROM ID on your BeagleBone should match one of these:
https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/master/adafruit_platformdetect/constants/boards.py#L291

pdp7:
@Cellgalvano please take a look at the proposed change in PR #440.  I don't have an original BeagleBone with me right now to test, but I think that should be what is needed.

Cellgalvano:
Hi,

thank you very much for the super fast response!
That is indeed the change I hoped for and for me it seems to work just fine. My board_id is set to 'BEAGLEBONE' as my Board is an original A6A Beaglebone and the detector got it right.

Best regards
Felix

pdp7:
@Cellgalvano good to hear.  is it ok to close the issue?

PR #440, opened 2021-04-05
Add support for the orignal BeagleBone #439

pdp7:
The original BeagleBone is same SoC (AM3358) as Black and Green and similar layout in terms of the cape headers.  It should just work once the EEPROM ID for original BeagleBone models is added this conditional logic.  Refer to issue #439


makermelissa (approved):
Code looks good to me. I didn't test because I don't have the board. :)

PR #452, opened 2021-05-03
Allwinner H6 added to digitalio.py

twa127:


makermelissa (approved):
Thanks, looks good to me.

Issue #454, opened 2021-05-04
Can't use an ST7789 TFT with Pico U2IF


2231puppy:
Hello, I can't seem to get my TFT(linked [here](https://www.adafruit.com/product/4313)) working with a Pi Pico running U2IF. It always gives me this error:

```
File "blinkatest.py", line 55, in <module>
    disp.image(image)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/adafruit_rgb_display/rgb.py", line 200, in image
    self._block(x, y, x + imwidth - 1, y + imheight - 1, pixels)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/adafruit_rgb_display/rgb.py", line 145, in _block
    self.write(self._RAM_WRITE, data)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/adafruit_rgb_display/rgb.py", line 297, in write
    spi.write(data)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/busio.py", line 339, in write
    return self._spi.write(buf, start, end)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/adafruit_blinka/microcontroller/pico_u2if/spi.py", line 49, in write
    pico_u2if.spi_write(buf, start=start, end=end)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/adafruit_blinka/microcontroller/pico_u2if/pico_u2if.py", line 326, in spi_write
    bytes([write_cmd, chunk]) + buffer[start : (start + chunk)], True
TypeError: can't concat list to bytes
```

I've tinkered around for ~3hrs after posting this problem in #446 and being asked to put it in a new issue. Not wanting to immediately do that, I messed with the code for a while until I turned to y'all, the Python wizards. The problem seems to be that the PIL object can't be converted to bytes, but if you use the raw image data as the bytes, it complains about it not having the right attributes. Might be totally wrong about that; just my best guess.

Thanks!

makermelissa:
I think the issue is the buffer is being passed in as a list instead of a bytestring. 

It can be reproduced in Python by attempting to concatenate a list of bytes to a bytestring:
```
>>> buffer = [0x23, 0x03, 0x65, 0x67, 0x34]
[...]

PR #455, opened 2021-05-04
Update pico_u2if.py

2231puppy:
Fixes an issue that causes the ST7789 driver not to work with Pico U2IF


makermelissa:
Fixes #454.

2231puppy:
Closing as the problem seems to be elsewhere.

Issue #453, opened 2021-05-04
Allwinner H6 pin.py


twa127:
Hi,

I think there may be an error in pin.py for the Allwinner H6, GPIO PL8 and PL9 are set as

PL8 = Pin((1, 360))
PL9 = Pin((1, 361))

The H6 appears to have 256 lines on GPIO Controller1, testing on my PineH64-B, I think the definition should be

PL8 = Pin((0, 8))
PL9 = Pin((0, 9))

makermelissa:
I think you're right. The AllWinner A64 also uses the same pin. Want to submit a PR?

makermelissa:
Fixed via #456

PR #456, opened 2021-05-05
Fix for Allwinner H6 PL8 and PL9 pin definitions, see #453 + Fix for Allwinner H616 SPI CLK naming

twa127:


makermelissa (approved):
Thanks, looks good.

PR #457, opened 2021-05-06
Add additional SPI ports for BCM2711

timonsku:
Currently the additional SPI ports on the Pi4 or CM4 are not usable in Blinka without doing this change manually.
SPI6 uses the same pins as the default SPI1 pins.

### DIFF
--- a/src/adafruit_blinka/microcontroller/bcm283x/pin.py
+++ b/src/adafruit_blinka/microcontroller/bcm283x/pin.py
@@ -142,6 +142,10 @@ spiPorts = (
     (0, SCLK, MOSI, MISO),
     (1, SCLK_1, MOSI_1, MISO_1),
     (2, SCLK_2, MOSI_2, MISO_2),
+    (3, D3, D2, D1), #SPI3 on Pi4/CM4
+    (4, D7, D6, D5), #SPI4 on Pi4/CM4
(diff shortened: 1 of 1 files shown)


makermelissa:
Are these hardware SPI ports?

makermelissa:
Never mind. Looks like there's a SPI6 as well: https://www.raspberrypi.org/documentation/hardware/raspberrypi/spi/README.md

I'm thinking maybe we should split out the BCM2711 from the BCM283x and point the Pi4 and CM4 boards to that.

timonsku:
Yea they are all hardware.
Agreed, in the long term that would be a good idea to split them. They are fairly different in terms of peripherals available.

timonsku:
https://elinux.org/RPi_BCM2711_GPIOs has a great table of all the alt modes

timonsku:
Whats your preferred way forward @makermelissa ?
Right now a release with this PR would help me a lot but I can understand if you want to split them right away.

makermelissa:
For now, let's go with this PR the way it is, but you'll need to run black and pylint to be sure it passes the failing checks.

timonsku:
Ran black and checked with pylint

makermelissa (approved):
Awesome, thanks.

PR #458, opened 2021-05-06
Set the I2C and SPI singletons conditionally if pins exist

### LINKED ISSUE #186: Should I2C / SPI singletons be provided conditionally?
The I2C and SPI singletons are currently setup regardless of whether they exists or not:
https://github.com/adafruit/Adafruit_Blinka/blob/738d6d910c2dfa9763c3f33ddf4692c4823e136f/src/board.py#L109-L117
They would then show up in:
```python
>>> import board
>>> dir(board)
```
Python allows those functions to be defined even though the pin variables do not exists. It would then fail when called:
```python
>>> SPI = board.SPI()
  File "board.py", line 120, in SPI
    return busio.SPI(SCLK, MOSI, MISO)
NameError: name 'SCLK' is not defined
>>> 
```

Should they be provided conditionally? I'm thinking a simple check for the existence of SCL/SDA pins for I2C and SCLK/MOSI/MISO pins for SPI.

For bitbang, that would require explicit creation from the `bitbangio` module, so these singletons aren't used in that scenario. (i think?)

makermelissa:
Fixes #186.

### DIFF
--- a/src/board.py
+++ b/src/board.py
@@ -224,15 +224,19 @@ elif "sphinx" in sys.modules:
 else:
     raise NotImplementedError("Board not supported {}".format(board_id))
 
-if SDA and SCL:
+try:
+    if SDA and SCL:
 
-    def I2C():
-        """The singleton I2C interface"""
-        return busio.I2C(SCL, SDA)
+        def I2C():
+            """The singleton I2C interface"""
+            return busio.I2C(SCL, SDA)
+
+    if SCLK:
+
+        def SPI():
+            """The singleton SPI interface"""
+            return busio.SPI(SCLK, MOSI, MISO)
 
 
-if SCLK:
-
-    def SPI():
-        """The singleton SPI interface"""
-        return busio.SPI(SCLK, MOSI, MISO)
+except NameError:
+    pass
--- a/src/board.py
+++ b/src/board.py
@@ -31,9 +31,9 @@ import sys
 
 import adafruit_platformdetect.constants.boards as ap_board
 from adafruit_blinka.agnostic import board_id, detector
-import busio
 
 # pylint: disable=wildcard-import,unused-wildcard-import,ungrouped-imports
+# pylint: disable=import-outside-toplevel
 
 if board_id == ap_board.FEATHER_HUZZAH:
     from adafruit_blinka.board.feather_huzzah import *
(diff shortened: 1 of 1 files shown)


caternuson:
I don't think it'll work to reference the pins directly in the conditional, since they won't exist. Here's MCP2221 as an example, which has no SPI port:
```python
(blinka):~$ export BLINKA_MCP2221=1
(blinka):~$ python3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
>>> import board
  File "/home/notacat/Adafruit_Blinka/src/board.py", line 235, in <module>
    if SCLK:
NameError: name 'SCLK' is not defined
>>> 
```

makermelissa:
Yeah, I need to use a try/except block. I'm just trying to decide if the If stuff is even necessary with that.

caternuson:
Maybe?
```python
>>> "SCLK" in locals()
False
>>>
```

makermelissa:
Tested on a Raspberry Pi (with the ports) and via sphinx (without the ports).

makermelissa:
Do you think I should move `import busio` back to where it was?

makermelissa:
Thinking about it, yeah. I probably should since we don't need it otherwise.

caternuson:
You sure the locals() trick won't work? I just did a quick hack check and it seemed to work.
```python
if "SCL" in locals():
    print("i can has i2c")

def I2C():
    """The singleton I2C interface"""
    import busio

    return busio.I2C(SCL, SDA)

if "SCLK" in locals():
    print("i can has spi")

def SPI():
    """The singleton SPI interface"""
    import busio

    return busio.SPI(SCLK, MOSI, MISO)

```

and then (again with MCP2221):
```python
>>> import board
i can has i2c
>>> 
```

You'd want to do more, but that just checks the basic logic plumbing works.

makermelissa:
Ok, I just tried the locals trick (I tested it incorrectly before) and that seemed to work. I can modify the code to use that.

caternuson:
The other thing is the defines don't actually throw any exceptions, despite the reference to undefined pins. It's only when you try to actually call them, like try to instantiate:
```python
>>> i2c = board.I2C()
```

Python let's you do this for some reason.
```python
>>> def foo():
...     print(I_DONT_EXIST)
... 
>>> foo()
  File "<stdin>", line 2, in foo
NameError: name 'I_DONT_EXIST' is not defined
>>> 
```

caternuson:
Looks good. Tested.

**FT232H**
```python
>>> import board
>>> dir(board)
['C0', 'C1', 'C2', 'C3', 'C4', 'C5', 'C6', 'C7', 'D4', 'D5', 'D6', 'D7', 'I2C', 'MISO', 'MOSI', 'SCK', 'SCL', 'SCLK', 'SDA', 'SPI', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'ap_board', 'board_id', 'detector', 'pin', 'sys']
>>> i2c = board.I2C()
>>> spi = board.SPI()
>>> 
```

**MCP2221**
```python
>>> import board
>>> dir(board)
['G0', 'G1', 'G2', 'G3', 'I2C', 'SCL', 'SDA', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'ap_board', 'board_id', 'detector', 'pin', 'sys']
>>> i2c = board.I2C()
>>> spi = board.SPI()
AttributeError: module 'board' has no attribute 'SPI'
>>> 
```

caternuson (approved):

PR #464, opened 2021-05-09
Updates for Allwinner H6

twa127:


makermelissa:
Pins are no longer needed for SPI because they were discarded anyways.

PR #468, opened 2021-05-10
Add Support for Rock Pi E

twa127:


makermelissa:
Do you still need to modify busio.py?

twa127:
oh dear, I'm having a bad day :), busio.py now modified

PR #474, opened 2021-05-20
Add missing iterations of rp2040 SPI0 Ports (GP18, GP19, GP16)

DollarStarNova:
we are missing the SPI 0 pins on the other side of the rp2040 (see port layout here - https://www.raspberrypi-spy.co.uk/wp-content/uploads/2021/01/raspberry_pi_pico_pinout.png)

I have added the missing pin configurations.


makermelissa:
Thanks, I'm not sure how I missed those. I may end up refactoring to have it automatically generate all the combos, but this will work for now.

DollarStarNova:
> Thanks, I'm not sure how I missed those. I may end up refactoring to have it automatically generate all the combos, but this will work for now.

Auto-generation would be a good idea eventually, yes :) As much as I enjoyed writing these by hand (any excuse for multiline cursors!) It's not the ideal solution.

makermelissa (approved):

PR #488, opened 2021-06-14
Fix docs not passing if microcontroller is imported

### LINKED ISSUE #487: GENERIC_X86 - Microcontroller not supported
Ref: https://github.com/adafruit/Adafruit_Blinka_Displayio/actions/runs/936869211 (if it's not been deleted)
PR: https://github.com/adafruit/Adafruit_Blinka_Displayio/pull/58
```
  File "/home/runner/work/Adafruit_Blinka_Displayio/Adafruit_Blinka_Displayio/displayio/i2cdisplay.py", line 30, in <module>
    import microcontroller
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/microcontroller/__init__.py", line 117, in <module>
    raise NotImplementedError("Microcontroller not supported:", chip_id)
NotImplementedError: ('Microcontroller not supported:', 'GENERIC_X86')
```

When running Github Actions, in the example above to build documentation, that uses `microcontroller` it will fail because there is not a fall-back case of `GENERIC_X86`.

Should this be added?
Or does this open up a whole can-of-worms that we don't want to get into?

In the example above, it can be fixed by removing the paramter typing that references `microcontroller.Pin`.

makermelissa:
Fixes #487.

### DIFF
--- a/src/microcontroller/__init__.py
+++ b/src/microcontroller/__init__.py
@@ -1,5 +1,6 @@
 """Microcontroller pins"""
 
+import sys
 import time
 
 from adafruit_platformdetect.constants import chips as ap_chip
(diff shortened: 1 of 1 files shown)


makermelissa:
Tested that this doesn't break MicroPython support.

lesamouraipourpre (approved):
Tested locally on Ubuntu Hirsuite Hippo and it works.

_Note: The I2CDisplay PR had been tested and pushed from Raspberry Pi 4 which was why it passed locally before._

PR #495, opened 2021-07-16
Added BeagleBone StarLight to Blinka (GPIO, I2C working)

makermelissa:
I added the new BeagleBone StarLight to Blinka and have GPIO and I2C working.
I also reduced the huge if/else tree in busio for SPI by making use of the any_embedded_linux property for anything else that's not caught ahead of that for loading up specialized constructors. That way we don't have an extra 20-30 if checks just to import the same line of code.

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -279,47 +279,7 @@ class SPI(Lockable):
 
     def configure(self, baudrate=100000, polarity=0, phase=0, bits=8):
         """Update the configuration"""
-        if detector.board.any_raspberry_pi or detector.board.any_raspberry_pi_40_pin:
-            from adafruit_blinka.microcontroller.generic_linux.spi import SPI as _SPI
-        elif detector.board.BEAGLEBONE_AI:
-            from adafruit_blinka.microcontroller.generic_linux.spi import SPI as _SPI
-        elif detector.board.any_beaglebone:
-            from adafruit_blinka.microcontroller.generic_linux.spi import SPI as _SPI
-        elif detector.board.any_orange_pi:
-            from adafruit_blinka.microcontroller.generic_linux.spi import SPI as _SPI
-        elif detector.board.any_nanopi and detector.chip.id == ap_chip.SUN8I:
-            from adafruit_blinka.microcontroller.generic_linux.spi import SPI as _SPI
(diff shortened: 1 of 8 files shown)


src/busio.py
```diff
@@ -279,47 +280,7 @@ def __init__(self, clock, MOSI=None, MISO=None):
 
     def configure(self, baudrate=100000, polarity=0, phase=0, bits=8):
         """Update the configuration"""
-        if detector.board.any_raspberry_pi or detector.board.any_raspberry_pi_40_pin:
```
ladyada, line 282: did u mean to delete?

ladyada (changes requested):

src/busio.py
makermelissa: Oh, we should add `any_raspberry_pi_40_pin` to https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/main/adafruit_platformdetect/board.py#L600-L624 first. It's not currently there, so it would stop working.

makermelissa: Submitted PR https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/184

ladyada (approved):

pdp7:
@makermelissa thanks!  please let me know if you run into any problems or have any questions.  Our newest Fedora image is July 7th which now uses 5.13 kernel.  We have set of patches upon mainline that myself and few others are working on getting upstreamed into Linux.  The Blinka support you added is helpful as I improve the GPIO and pinctrl kernel drivers I can have something to test against for regressions.

makermelissa:
> @makermelissa thanks! please let me know if you run into any problems or have any questions. Our newest [Fedora image is July 7th which now uses 5.13 kernel](https://github.com/starfive-tech/Fedora_on_StarFive#download-the-latest-image). We have set of patches upon mainline that myself and few others are working on getting upstreamed into Linux. The Blinka support you added is helpful as I improve the GPIO and pinctrl kernel drivers I can have something to test against for regressions.

Awesome. That's the image I used when I added support. I wasn't able to get SPI working, but I do run into that pretty often with many boards. My biggest concern is that in future images the /proc/device-tree/compatible will change enough that it won't be detected anymore, but it's also an easy fix.

PR #498, opened 2021-07-27
Added keypad module

makermelissa:
This is a complete implementation.

### DIFF
--- a/src/keypad.py
+++ b/src/keypad.py
@@ -1,0 +1,488 @@
+"""
+`keypad` - Support for scanning keys and key matrices
+===========================================================
+See `CircuitPython:keypad` in CircuitPython for more details.
+
+* Author(s): Melissa LeBlanc-Williams
+"""
+
+import time
+import threading
+from collections import deque
+import digitalio
+
+
+class Event:
+    """A key transition event."""
+
+    def __init__(self, key_number=0, pressed=True):
+        """
+        Create a key transition event, which reports a key-pressed or key-released transition.
+
+        :param int key_number: the key number
+        :param bool pressed: ``True`` if the key was pressed; ``False`` if it was released.
+        """
+        self._key_number = key_number
+        self._pressed = pressed
+
+    @property
+    def key_number(self):
+        """The key number."""
+        return self._key_number
+
+    @property
+    def pressed(self):
+        """
+        ``True`` if the event represents a key down (pressed) transition.
+        The opposite of `released`.
+        """
+        return self._pressed
+
+    @property
+    def released(self):
+        """
+        ``True`` if the event represents a key up (released) transition.
+        The opposite of `pressed`.
+        """
+        return not self._pressed
+
+    def __eq__(self, other):
+        """
+        Two `Event` objects are equal if their `key_number`
+        and `pressed`/`released` values are equal.
+        """
+        return self.key_number == other.key_number and self.pressed == other.pressed
+
+    def __hash__(self):
+        """Returns a hash for the `Event`, so it can be used in dictionaries, etc.."""
+        return hash(self._key_number)
+
+    def __repr__(self):
+        """Return a textual representation of the object"""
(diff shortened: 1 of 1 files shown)


ladyada (approved):

src/keypad.py
```diff
@@ -0,0 +1,488 @@
...
+        """
+        Automatically deinitializes when exiting a context. See
+        :ref:`lifetime-and-contextmanagers` for more info.
+        """
+        self.deinit()
+
+    def _scanning_loop(self):
+        while True:
+            self._scanning_function()
+            time.sleep(0.001)
```
dhalbert, line 184: Suppose you just call the scanning function once per interval here? Then you don't need to check in each internal scanning function. In the C implementation, this was not as easy, but it's easier here.
```suggestion
            time.sleep(self._interval)
```

src/keypad.py
```diff
@@ -0,0 +1,488 @@
...
+        if len(self._events) == self._events.maxlen:
+            self._overflowed = True
+        else:
+            self._events.append(Event(key_number, current))
+
+
+class _KeysBase:
+    def __init__(self, interval, max_events, scanning_function):
+        self._interval = interval
+        self._last_scan = time.monotonic()
```
dhalbert, line 154: If checking interval in thread loop, as suggested below:
```suggestion
```

src/keypad.py
```diff
@@ -0,0 +1,488 @@
...
+    @property
+    def key_count(self):
+        """The number of keys that are being scanned. (read-only)"""
+        return len(self._digitalinouts)
+
+    def _keypad_keys_scan(self):
+        if time.monotonic() - self._last_scan < self._interval:
+            return
+        self._last_scan = time.monotonic()
+
```
dhalbert, line 257: If checking interval in thread loop, as suggested above:
```suggestion
```

src/keypad.py
```diff
@@ -0,0 +1,488 @@
...
+        self._previously_pressed = self._currently_pressed = [False] * self.key_count
+
+    def _row_column_to_key_number(self, row, column):
+        return row * len(self._column_digitalinouts) + column
+
+    def _keypad_keymatrix_scan(self):
+        if time.monotonic() - self._last_scan < self._interval:
+            return
+        self._last_scan = time.monotonic()
+
```
dhalbert, line 354: If checking interval in thread loop, as suggested above:
```suggestion
```

src/keypad.py
```diff
@@ -0,0 +1,488 @@
...
+    @property
+    def events(self):
+        """The `EventQueue` associated with this `Keys` object. (read-only)"""
+        return self._events
+
+    def _keypad_shiftregisterkeys_scan(self):
+        if time.monotonic() - self._last_scan < self._interval:
+            return
+        self._last_scan = time.monotonic()
+
```
dhalbert, line 477: If checking interval in thread loop, as suggested above:
```suggestion
```

dhalbert (changes requested):
Nice and clean! What do you think of my idea of sleep for the interval in the thread loop?

src/keypad.py
makermelissa: Though this simplifies things, it loses accuracy because it is sleep interval + processing time. I could probably do something more accurate like grabbing time.sleep(self._interval - (time.monotonic - self._last_scan)).

dhalbert: I was thinking this would use more CPU (polling the scanning routines every msec instead of, say, every 20msecs). We are talking about msec or 10-msec granularity, so I'm not too worried about accuracy. But it's not a big deal -- I'm happy to  approve as is.

makermelissa: True that it's just a tiny sliver of time, but I do like to make it as accurate as possible :) Anyway, I updated in the spirit of your suggestion.

dhalbert (approved):
Thanks, and nice!

PR #501, opened 2021-08-15
mcp2221: allow empty i2c payload

fgervais:
This fix allow sending I2C transactions with an empty payload.

Before this change, sending an empty payload is bogus as it won't transfer
anything but will still try to get the status of the transaction. It will
basically get the current status of the bus.

One way to manifest this problem is when connecting to a device.

On `I2CDevice` init, it will [probe the device by sending en empty payload](https://github.com/adafruit/Adafruit_CircuitPython_BusDevice/blob/b869a4a12618721f1232cbbec95a6186fb84ddf7/adafruit_bus_device/i2c_device.py#L154)
message. This actually works fine if the bus is in a good state like on
first use. It won't actually send the probe out but we will read the
good initial state and things will go smoothly from here.

However if we start by initializing a device with an address at which
there is no device, it won't send the empty payload probe but will fail
later on as there is no device at that address.

At this point the bus is basically in DoS as trying to init a device
will again not send the empty payload but now since the bus is in the
error state it will fail even if the device is in fact present.

### DIFF
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -198,7 +198,7 @@ class MCP2221:
         length = end - start
         retries = 0
 
-        while (end - start) > 0:
+        while (end - start) > 0 or not buffer:
             chunk = min(end - start, MCP2221_MAX_I2C_DATA_LEN)
             # write out current chunk
             resp = self._hid_xfer(
@@ -223,6 +223,8 @@ class MCP2221:
             # yay chunk sent!
             while self._i2c_state() == RESP_I2C_PARTIALDATA:
                 time.sleep(0.001)
+            if not buffer:
+                break
             start += chunk
             retries = 0
 
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -189,7 +189,7 @@ class MCP2221:
             # bus release will need "a few hundred microseconds"
             time.sleep(0.001)
 
-    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-arguments,too-many-branches
     def _i2c_write(self, cmd, address, buffer, start=0, end=None):
         if self._i2c_state() != 0x00:
             self._i2c_cancel()


caternuson:
Finally got around to looking at this. Was able to recreate what I think is the same issue using an AW9523:
```python
(blinka) user@computer:~$ python3
Python 3.8.10 (default, Jun  2 2021, 10:49:15) 
[GCC 9.4.0] on linux
>>> import board
>>> import adafruit_aw9523
>>> i2c = board.I2C()
>>> i2c.scan()
[88]
>>> aw = adafruit_aw9523.AW9523(i2c)
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_aw9523.py", line 71, in __init__
    self.i2c_device = i2c_device.I2CDevice(i2c_bus, address)
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_bus_device/i2c_device.py", line 50, in __init__
    self.__probe_for_device()
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_bus_device/i2c_device.py", line 154, in __probe_for_device
    self.i2c.writeto(self.device_address, b"")
  File "/home/user/blinka/lib/python3.8/site-packages/busio.py", line 158, in writeto
    return self._i2c.writeto(address, buffer, stop=stop)
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_blinka/microcontroller/mcp2221/i2c.py", line 19, in writeto
    self._mcp2221.i2c_writeto(address, buffer, start=start, end=end)
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_blinka/microcontroller/mcp2221/mcp2221.py", line 315, in i2c_writeto
    self._i2c_write(0x90, address, buffer, start, end)
  File "/home/user/blinka/lib/python3.8/site-packages/adafruit_blinka/microcontroller/mcp2221/mcp2221.py", line 233, in _i2c_write
    raise RuntimeError("I2C slave address was NACK'd")
RuntimeError: I2C slave address was NACK'd
>>> 
```

With this PR:
```python
(blinka) user@computer:~$ python3
Python 3.8.10 (default, Jun  2 2021, 10:49:15) 
[GCC 9.4.0] on linux
>>> import board
>>> import adafruit_aw9523
>>> i2c = board.I2C()
>>> i2c.scan()
[88]
>>> aw = adafruit_aw9523.AW9523(i2c)
>>> aw.inputs
0
>>> aw.inputs
2
>>>  
```
(pin 1 connected to Vin on second read)

@fgervais Thanks for the fix!

LGTM so merging.

PR #505, opened 2021-09-10
Update digitalio docstrings to be more in line with CircuitPython

### LINKED ISSUE #499: misleading docstr in DigitalInOut.pull
https://github.com/adafruit/Adafruit_Blinka/blob/4f4e6af0a2c2912d7662f0160849806507b125bd/src/digitalio.py#L190-L191

I think this docstr should say
```py
"""Get or set the Digital Pin's input impedance."""
```

ps - I think this is simply a copy-n-paste mistake.

makermelissa:
Fixes #499.

### DIFF
--- a/src/adafruit_blinka/microcontroller/generic_linux/spi.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/spi.py
@@ -129,7 +129,7 @@ class SPI:
             self._spi.bits_per_word = self.bits
             data = self._spi.transfer(list(buffer_out[out_start : out_end + 1]))
             for i in range((in_end - in_start)):
-                buffer_in[i + in_start] = data[i + 1]
+                buffer_in[i + in_start] = data[i]
             # self._spi.close()
         except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


src/adafruit_blinka/microcontroller/generic_linux/spi.py
```diff
@@ -129,7 +129,7 @@ def write_readinto(
             self._spi.bits_per_word = self.bits
             data = self._spi.transfer(list(buffer_out[out_start : out_end + 1]))
             for i in range((in_end - in_start)):
-                buffer_in[i + in_start] = data[i]
+                buffer_in[i + in_start] = data[i + 1]
```
FoamyGuy, line 132: The docs changes look good to me.

Is this `+ 1` intentional? If so this PR  LGTM.

makermelissa: Sort of. That wasn't supposed to be part of this PR. It was for another issue that I started to test a while ago and forgot to revert. I'll remove that for now.

PR #508, opened 2021-09-20
Add versioning to the top-level Python files

lesamouraipourpre:
Add version munging to the top-level Python files so that when a new version is released the files uploaded to PyPI will have a `__version__` and `__repo__` variable similar to the CircuitPython ones.

eg, `busio.__version__ = 6.14.0`

There is also a special variable only added to `board` for now:
`board.__blinka__ = True`
This is to make it easy for a library to determine if it is being run on Blinka.

Tested by uploading a dummy release to:
https://pypi.org/project/samourai-Adafruit-Blinka/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,5 +29,8 @@ jobs:
         TWINE_USERNAME: ${{ secrets.pypi_username }}
         TWINE_PASSWORD: ${{ secrets.pypi_password }}
       run: |
+        for file in $(find -not -path "./.*" -not -path "./docs*" -name "*.py"); do
+            sed -i -e "s/0.0.0-auto.0/${{github.event.release.tag_name}}/" $file;
+        done;
         python setup.py sdist
         twine upload dist/*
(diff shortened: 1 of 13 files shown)


makermelissa (approved):
Thank you! Also, thanks for adding in keypad. I apparently missed that when I wrote it.

makermelissa:
Ok @lesamouraipourpre, I thought about this a bit more overnight and realized that the CircuitPython modules actually don't have versions. For instance:
```
Adafruit CircuitPython 7.0.0 on 2021-09-20; Adafruit Feather M4 Express with samd51j19
>>> import busio
>>> busio.__version__
AttributeError: 'module' object has no attribute '__version__'
```
We currently have a `__version__` for Blinka (https://github.com/adafruit/Adafruit_Blinka/blob/main/src/__version__.py), but it's not automatically updated and appears to still say 1.0.0, so it's basically useless. I'm thinking about rolling back this PR and I think we should implement the version munging into there if that's possible. Let me know what you think before I do roll it back.

lesamouraipourpre:
You're right, my brain defaulted to thinking of libraries instead of the core.

I couldn't figure out any way of using the `__version__.py` file from Python/Blinka.
I would propose keeping the `__version__` in `board` so that we can determine the version of Blinka at runtime.

makermelissa:
Ok, I'll roll back the changes and then perhaps you can create a new one with just board. I don't think there's a way to only do a partial rollback.

PR #511, opened 2021-09-21
Document the examples

lesamouraipourpre:
Add the examples that exist to the documention.
*Note: I've not checked these for correctness*

Remove the execute bits from the examples that don't start '#!/'


makermelissa (approved):
Thanks

PR #522, opened 2021-10-20
Split RP4 board definitions

AdamCummick:
Ran into an issue using the SPI1 bus where I needed CPOL=1. On the Raspberry Pi 4 SPI1 does not support CPOL=1, however SPI6 has identical pinout to SPI1, aside from fewer hardware CS pins, and is fully featured. 

Change checks which board is being used to determine SPI pins.

If this is merged, the [learning guide](https://learn.adafruit.com/circuitpython-on-raspberrypi-linux/spi-sensors-devices) should also be updated as the newer Pi's would require a different config.txt setting: `dtoverlay=spi6-2cs`


makermelissa:
Ok cool, it's passing. I'd like to test this PR out soon before we merge it.

PR #528, opened 2021-11-29
change actions to use python 3.7

FoamyGuy:
This updates the version of python that will get used in the actions container to 3.7. This is needed to avoid an issue currently affecting debian builds. It's discussed a bit more here: https://github.com/adafruit/Adafruit_Blinka_Displayio/pull/73

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,10 +13,10 @@ jobs:
     - name: Translate Repo Name For Build Tools filename_prefix
       id: repo-name
       run: echo ::set-output name=repo-name::Adafruit-Blinka
-    - name: Set up Python 3.6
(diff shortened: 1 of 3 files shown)


makermelissa (changes requested):
Can you also change https://github.com/adafruit/Adafruit_Blinka/blob/main/setup.py#L48, https://github.com/adafruit/Adafruit_Blinka/blob/main/setup.py#L83, and https://github.com/adafruit/Adafruit_Blinka/blob/main/docs/conf.py#L32?

FoamyGuy:
Newest commit updates the two spots in setup.py and the one in docs/conf.py

makermelissa (approved):
Thank you

Issue #525, opened 2021-11-07
Support for BeagleBone Green Gateway


keenanjohnson:
I see that the [Beaglebone Green Gateway](https://www.seeedstudio.com/SeeedStudio-BeagleBone-Green-Gateway-p-4586.html) is not currently supported by this library. I think it might be nice for this board to be supported as it has some great features!

Does it seem reasonable to support this board here?

keenanjohnson:
It seems straightforward to add this to the board.py file since the BEAGLEBONE_GREEN is already supported and the pin assignments will be the same.

https://github.com/adafruit/Adafruit_Blinka/blob/f8c28157c2ffceaa436f106f0bbc4678b889cc39/src/board.py#L82-L83

I'm going to go for it!

makermelissa:
Fixed via #529

PR #529, opened 2021-12-06
#525 Add Beaglebone Green Gateway Board Support

keenanjohnson:
Hello Adafruit Blinka maintainers team! 

I have a [Beaglebone Green Gateway board](https://www.seeedstudio.com/SeeedStudio-BeagleBone-Green-Gateway-p-4586.html) that I would like to use Blinka on! It seems trivial to add it, so I've added support here using the same pattern that is used to add support for the Beaglebone Green!

I have submitted my change here. Let me know if this looks good or if anything needs changing!

### DIFF
--- a/src/board.py
+++ b/src/board.py
@@ -82,6 +82,9 @@ elif board_id == ap_board.BEAGLEBONE_BLACK:
 elif board_id == ap_board.BEAGLEBONE_GREEN:
     from adafruit_blinka.board.beagleboard.beaglebone_black import *
 
+elif board_id == ap_board.BEAGLEBONE_GREEN_GATEWAY:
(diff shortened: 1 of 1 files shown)


makermelissa:
Hi, thanks for adding this. You didn't mention if you had tested it yet and which features you tested. Thanks.

makermelissa (approved):
Actually, this is fine. Since it just piggybacks on the BB Green, I'm just going to add it.

keenanjohnson:
Thanks for merging this! Yes I forgot to mention testing, but I had built the library locally on my beaglebone green gateway and verified that the I2C libraries work as I had expected! Thanks! @makermelissa!

PR #533, opened 2021-12-18
add case for generic_x86

FoamyGuy:
Allow successful import on `GENERIC_X86`. Warn that all features may not work.


makermelissa (approved):
Looks good to me.

Issue #536, opened 2021-12-31
Improvements to the `mcp2221.py` module


ezio-melotti:
As a follow-up to #535, I'm doing more tests and tweaks to the [mcp2221.py](https://github.com/adafruit/Adafruit_Blinka/blob/main/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py) file, in particular:
1. I've added `self._hid.close()` in a few places to [ensure the device is closed correctly](https://trezor.github.io/cython-hidapi/api.html#hid.device.close), avoiding segfaults upon exit.
2. I'm trying to make the native `hid_mpc2221` driver work out of the box, without having to replace it with the`hid_generic` driver through `rmmod`/blacklisting:

   * Changing `5` to `20` at line 123 seems to solve the `OSError: open failed` reported in #535, without having to rely on a forced `sleep()` at line 121: https://github.com/adafruit/Adafruit_Blinka/blob/8a25ee484e7aed8c58e4366b3fca78083bc61422/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py#L119-L125
     I would replace the hardcoded `5` with `MCP2221_RESET_DELAY`, remove the `sleep`, and increase the default value of `MCP2221_RESET_DELAY` to 20/30s.
   * Even though the error is solved, the `open` at line 125 still takes several seconds with the native `hid_mpc2221` driver.  I tried to close the device before the `open` and also to create a separate device instance, but it's always slow.  This seems to be caused by the reset, but I couldn't find any documentation about that reset byte sequence, nor a way to make it fast.  A minimal reproducer is:
     ```py
     >>> # open the device
     >>> h = hid.device(); h.open(0x04D8, 0x00DD)
     >>> # close and reopen -- it's fast
     >>> h.close(); print('closed'); h.open(0x04D8, 0x00DD)
     closed
     >>> # send the reset bytestring
     >>> report = b"\x70\xAB\xCD\xEF"; h.write(b"\0" + report + b"\0" * (64 - len(report)))
     65
     >>> # close returns immediately, but now open fails
     >>> h.close(); print('closed'); h.open(0x04D8, 0x00DD)
     closed
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
       File "hid.pyx", line 113, in hid.device.open
     OSError: open failed
     >>> # after ~16s open works again
     >>> h.close(); print('closed'); h.open(0x04D8, 0x00DD)
     closed
     >>>
     ```
[...]

PR #540, opened 2022-01-06
Close the mcp2221 device before exit.

ezio-melotti:
This PR addresses the first point of #536.  The [`hidapi` docs](https://trezor.github.io/cython-hidapi/api.html#hid.device.close) mention explicitly that `device.close()` "should always be called after opening a connection".

To address this issue, I first created a new public `MCP2221.close()` method.  This has three advantages:
1. the user can close the device directly;
2. the method can be used as a callback to close the currently-open device (using `atexit.register(self._hid.close)` doesn't work if `self._hid` gets reassigned);
3. the class becomes usable as a context manager through `contextlib.closing` (even though it might be better to explicitly support the context-manager protocol by adding `__enter__` and `__exit__`).

I then called this method in 3 places, to ensure the device is closed:
1. In a `try/except` around the `self._reset()` call in the `__init__`;
2. In `MCP2221.__del__`, called when the instance is destroyed (note that this is not guaranteed);
3. on program exit (using `atexit`).

Calling `.close()` on an already closed device does nothing, so the extra redundancy that might cause multiple calls to `.close()` shouldn't be an issue.

### DIFF
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -55,14 +55,7 @@ class MCP2221:
         # make sure the device gets closed before exit
         atexit.register(self.close)
         if MCP2221_RESET_DELAY >= 0:
-            try:
-                self._reset()
-            except OSError:
-                # this might fail on Linux with the hid_mcp2221 native
-                # driver and a short reset delay -- if it fails,
-                # close the device before reraising
-                self.close()
-                raise
+            self._reset()
         self._gp_config = [0x07] * 4  # "don't care" initial value
         for pin in range(4):
             self.gp_set_mode(pin, self.GP_GPIO)  # set to GPIO mode
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -2,6 +2,7 @@
 
 import os
 import time
+import atexit
 import hid
 
 # Here if you need it
@@ -51,12 +52,29 @@ class MCP2221:
     def __init__(self):
         self._hid = hid.device()
         self._hid.open(MCP2221.VID, MCP2221.PID)
+        # make sure the device gets closed before exit
+        atexit.register(self.close)
         if MCP2221_RESET_DELAY >= 0:
-            self._reset()
+            try:
+                self._reset()
+            except OSError:
+                # this might fail on Linux with the hid_mcp2221 native
+                # driver and a short reset delay -- if it fails,
+                # close the device before reraising
+                self.close()
+                raise
         self._gp_config = [0x07] * 4  # "don't care" initial value
         for pin in range(4):
             self.gp_set_mode(pin, self.GP_GPIO)  # set to GPIO mode
             self.gpio_set_direction(pin, 1)  # set to INPUT
+
+    def close(self):
+        """Close the device. Does nothing if the device is not open."""
+        self._hid.close()
+
+    def __del__(self):
+        # try to close the device before destroying the instance
+        self.close()
 
     def _hid_xfer(self, report, response=True):
         """Perform HID Transfer"""
--- a/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
+++ b/src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
@@ -3,6 +3,7 @@
 import os
 import time
 import atexit
+
 import hid
 
 # Here if you need it
@@ -62,7 +63,7 @@ class MCP2221:
             self.gpio_set_direction(pin, 1)  # set to INPUT
 
     def close(self):
-        """Close the device. Does nothing if the device is not open."""
+        """Close the hid device. Does nothing if the device is not open."""
         self._hid.close()
 
     def __del__(self):


caternuson:
Have you run into a use case that is having issues with the current driver code? Or is this just based on the comment in the hidapi docs?

ezio-melotti:
Leaving the device open causes a segfault upon exit while using the `hid_mcp2221` driver.

As part of #536, I'm trying to make the mcp2221 work (almost) out of the box on Linux using the native `hid_mcp2221` driver without having to blacklist it and replace it with the `hid_generic` one.  The segfault is one of the issues, and closing the device fixes that and shouldn't have downsides on other environments.

caternuson:
So it's specific to use with `hid_mcp2221` driver?

I did all the dev work and testing for the current code on linux and never saw seg faults. So I figured there was some final cleanup (close) being done in `hid_api`.

ezio-melotti:
Yes, it seems that the `hid_generic` driver is able to handle open devices more gracefully upon exit, whereas the `hid_mcp2221` crashes (sometimes with a segfault, sometimes with a failed assertion).

dhalbert:
@caternuson Just checking - what's the status of this?

caternuson:
I suggest we close this. The issue is being seen using a different underlying driver than the current documentation shows. The lack of any kind of clean up and calling `close()` _may_ be an issue, but it needs to be demonstrated with the currently documented setup. Then we can investigate.

ezio-melotti:
The current setup using the `hid_generic` driver seems to work even without calling `close()`.  Calling `close()` doesn't seem to cause any problem either, so the situation shouldn't change much whether the PR is merged or not.

With the `hid_mcp2221` driver the situation is a bit different because not calling `close()` causes a crash.  However the crash is both good and bad: bad because the crash shouldn't happen in the first place, good because there is currently a bug with the `hid_mcp2221` driver that might leave the MCP2221 in a "broken" state, and the crash resets the state.

I've been in touch with the author of the `hid_mcp2221` driver (@RishiGupta12) and he has [a patch that might fix the error](https://lore.kernel.org/all/20210818152743.163929-2-tobias.junghans@inhub.de/t/), but I haven't been able to test it yet.  Once the driver is fixed in the kernel and this PR is merged, calling close will ensure that everything will work with both drivers (and simplify the documented setup by simply using the default `hid_mcp2221` driver instead of blacklisting it and replace it with the `hid_generic`).  #536 has more details about this.

caternuson:
OK, maybe worth doing something here. Was doing some other testing and found a way to generally demonstrate an issue using the `hidapi` module. It's on a Raspberry Pi, which is an odd use case for MCP2221.

Here's a simple test:

**EXIT WITHOUT CLOSING**
```python
pi@raspberrypi:~ $ python3
Python 3.9.2 (default, Mar 12 2021, 04:06:34) 
[GCC 10.2.1 20210110] on linux
>>> import hid
>>> device = hid.device()
>>> device.open(0x04D8, 0x00DD)
>>> exit()
python3: ../../libusb/os/threads_posix.h:58: usbi_mutex_destroy: Assertion `pthread_mutex_destroy(mutex) == 0' failed.
python3: ../../libusb/os/threads_posix.h:46: usbi_mutex_lock: Assertion `pthread_mutex_lock(mutex) == 0' failed.
Aborted
pi@raspberrypi:~ $ 
```

**EXIT AFTER CLOSING**
```python
pi@raspberrypi:~ $ python3
Python 3.9.2 (default, Mar 12 2021, 04:06:34) 
[GCC 10.2.1 20210110] on linux
>>> import hid
>>> device = hid.device()
>>> device.open(0x04D8, 0x00DD)
>>> device.close()
>>> exit()
pi@raspberrypi:~ $ 
```

Similarly, can create same with a basic Blinka usage:
```python
pi@raspberrypi:~ $ python3
Python 3.9.2 (default, Mar 12 2021, 04:06:34) 
[GCC 10.2.1 20210110] on linux
>>> import board
>>> exit()
python3: ../../libusb/os/threads_posix.h:58: usbi_mutex_destroy: Assertion `pthread_mutex_destroy(mutex) == 0' failed.
python3: ../../libusb/os/threads_posix.h:46: usbi_mutex_lock: Assertion `pthread_mutex_lock(mutex) == 0' failed.
Aborted
pi@raspberrypi:~ $ 
```

This behavior does **not** exist on an x86 ubuntu setup, which is where all the original dev work was done.
```python
linux:~$ python3
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
>>> import board
>>> exit()
linux:~$
```

These don't seem to currently be causing any subsequent issues with any of the other usage outlined in the MCP2221 guide. Everything generally works OK.

src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
```diff
@@ -51,13 +52,30 @@ class MCP2221:
...
         if MCP2221_RESET_DELAY >= 0:
-            self._reset()
+            try:
+                self._reset()
+            except OSError:
+                # this might fail on Linux with the hid_mcp2221 native
+                # driver and a short reset delay -- if it fails,
+                # close the device before reraising
+                self.close()
+                raise
```
caternuson, line 65: Please revert this part back to original reset code since it seems specific to the `hid_mcp2221` module. The reset delay can be increased via an environment variable `BLINKA_MCP2221_RESET_DELAY`.

caternuson (changes requested):

src/adafruit_blinka/microcontroller/mcp2221/mcp2221.py
ezio-melotti: I just tested again with the `hid_mcp2221` driver and the comment is indeed a bit misleading.  I checked that:
* sending the reset signal just after opening works without errors, and closes the device, so the `close()` after this point is not necessary:
  ```py
  >>> import hid; h = hid.device()
  >>> h.open(0x04D8, 0x00DD); report = b"\x70\xAB\xCD\xEF"; h.write(b"\0" + report + b"\0" * (64 - len(report)))
  65
  >>>
  ```
* trying to open the device again just after the reset signal fails with an `OSError` if there isn't enough delay, however the device is still considered closed, and if the `open()` is not successful the call to `.close()` is not necessary.

This `try`/`except` would still be useful is in case sending the reset signal fails, raising an error and leaving the device open, but I haven't encountered this situation.  Calling `.close()` on a closed device is a no-op though, so I initially added it just as an extra precaution.

Let me know if you still want me to revert the change or just update the comment to explain that the `self.close()` call is just there to guarantee that the device is indeed closed before re-raising.

caternuson (approved):

caternuson:
Awesome. Thanks. Let's merge this!

ezio-melotti:
Thanks for the reviews and for merging this!  Do you want me to open another PR about 2.i of #536?

That will achieve two things:
1. the `reset` call will return as soon as the device can be reopened, without having to wait for the full reset delay to be passed
2. this will allow to increasing the default reset delay so that the MCP will have more time and won't fail with `OSError: open failed`

PR #543, opened 2022-01-19
Add typing module(s) to Blinka

tekktrik:
Address inability to use CircuitPython's typing module `_typing` (which will become `circuitpython_typing`) by adding a subset of the module.  In CircuitPython these are stub files, as my understanding is that these are purely conceptual (`bytearray` of course is a real thing, but a `ReabableBuffer` isn't and is just a way of communicating things that implement the buffer reading protocol).

This will works for CircuitPython, because adding the too the `try`/`except` block were using for typing in libraries means that it will never actually try to import the library, while the computer environment used to write code **will** be able to and will access the stubs.  However, since the `try`/`block` is intended to fail by importing the CPython `typing` library, platforms running Blinka won't actually fail, so my understanding is that these must be actual modules (`.py` instead of `.pyi`).

The only subsets of the CircuitPython typing modules implemented here are ones that could actually be used within an environment running Blinka, so they either have CPython implementations or CPython equivalents.  The latter translates to allowing the libraries that CircuitPython was previously emulating (e.g. - allowing the actual CPython `array.array` library and NumPy's `numpy.ndarray` that `ulab.numpy.array` was emulating).

It's worth mentioning that the latter example of import a NumPy class for typing may eat up memory that never actually gets used functionally in code, so it's worth asking if it's worth doing so for the sake of completeness.  But I also don't know too much about typical Blinka use so I don't know how catastrophic this is.

Happy to make changes as needed!

Here's how CircuitPython defines the typings:
https://github.com/adafruit/circuitpython/blob/main/shared-bindings/circuitpython_typing/__init__.pyi

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4,8 +4,11 @@
 .. If your library file(s) are nested in a directory (e.g. /adafruit_foo/foo.py)
 .. use this format as the module name: "adafruit_foo.foo"
 
+.. automodule:: _typing
+  :members
+
 .. automodule:: adafruit_blinka
-   :members:
+  :members:
 
 .. automodule:: adafruit_blinka.agnostic
   :members:
@@ -23,6 +26,9 @@
   :members:
 
 .. automodule:: busio
+  :members:
+
+.. automodule:: circuitpython_typing
   :members:
 
 .. automodule:: digitalio
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,7 +5,7 @@
 .. use this format as the module name: "adafruit_foo.foo"
 
 .. automodule:: _typing
-  :members
+  :members:
 
 .. automodule:: adafruit_blinka
   :members:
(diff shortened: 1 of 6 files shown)


dhalbert:
@tekktrik are you using `pre-commit` locally? It might save a bunch of testing round-trips.

dhalbert:
https://learn.adafruit.com/creating-and-sharing-a-circuitpython-library/check-your-code

tekktrik:
Yeah, but most of what I'm failing on is building documentation, I could do that instead of there's a good way to do it like the CI.  Caveat is I use Windows, if that matters.

dhalbert:
You can run sphinx locally, just copy what you see in the tests:
```
sphinx-build -E -W -b html . _build/html
```
You can install sphinx with pip.

tekktrik:
Alright we finally got there haha

makermelissa (approved):
Looks good. Thanks. We should probably also update the Blinka Displayio to use these instead.

tekktrik:
Thanks y'all!

makermelissa:
Thank you!

PR #544, opened 2022-01-19
Typing Library Documentation Touchups

tekktrik:
Companion PR of #543!

1. Adds typing libraries to README so they show up in repo and documentation landing
2. Swap author and note so it doesn't look like something is missing if you go to the `_typing` documentation
3. Fix bullet point list that isn't showing properly currently


makermelissa (approved):
Looks good. Thanks.

PR #548, opened 2022-02-10
Set executable flag on libgpiod_pulsein64 so pip installs it

makermelissa:
Fixes #547.

PR #549, opened 2022-02-10
Fix libgpiod further by updating setup.py

makermelissa:
After testing #548 and seeing that didn't fix it, I found out the file also needs to be added to setup.py, so this PR does that.


makermelissa:
Any chance this could get a quick look?

kattni (approved):
LGTM

PR #552, opened 2022-02-11
Remove numpy import

tekktrik:
Hotfixes issues with `adafruit_bus_device` and other libraries that are importing `circuitpython_typing` on Blinka platforms!


FoamyGuy (approved):
Looks good to me. Thank you @tekktrik 

Will merge after CI passing.

I'll tinker with it some over the weekend to see what affect it has on the docs, if there is substantial upside and folks prefer it we can eventually add it back with the addition to requirements.txt.

PR #554, opened 2022-02-13
remove PWMOut from pulseio

### LINKED ISSUE #503: Create pwmio to match core CircuitPython
`PWMOut` was split out from `pulseio` into `pwmio` with the release of CircuitPython 6.0

Should we do the same in Blinka so that any libraries that expect the `pwmio` module will work?

dhalbert:
- Fixes #503.
---
- Remove `pulseio.PWMOut` imports
- Move `bcm283x` implementation of `PWMOut` to `pwmio` directory.


ladyada:
coral boards do have true PWM out, this doesnt remove support right?
https://learn.adafruit.com/circuitpython-on-google-coral-linux-blinka/pwm-outputs-servos

dhalbert:
> coral boards do have true PWM out, this doesnt remove support right? https://learn.adafruit.com/circuitpython-on-google-coral-linux-blinka/pwm-outputs-servos

The PR is just removing `pulseio.PWMOut`, which was dropped in CPy 7. It leaves `pwmio.PWMOut` as is.

makermelissa:
Did we want to remove pulseio completely from here? I noticed it's still there for the Raspberry Pi.

dhalbert:
> Did we want to remove pulseio completely from here? I noticed it's still there for the Raspberry Pi.

It provides `pulseio.PulseIn`, which is real, I thought. So it's still there, but just for that.

makermelissa (approved):
Ok, sounds good to me.

dhalbert:
I did check the Learn Guide code for uses of `pulseio.PWMOut`, and didn't find any, which is good.

PR #556, opened 2022-02-15
remove _typing and circuitpython_typing

dhalbert:
`circuitpython_typing` is now a pypi library of its own (`adafruit-circuitpython-typing`), and no longer needs to be part of `adafruit-blinka`.

Remove the typing source files, and all mentions.

I don't thinking anything using _typing now. I think we can discard that.

I also ran pre-commit experimentally. I didn't pay attention to the `pylint` results, but it did make one `black` change and added a missing newline at the end of a file.

Don't merge this until these two libraries are updated to require `adafruit-circuitpython-typing`:
- [x] `debug-i2c`
- [x] `adafruit_bus_device`


dhalbert:
I suggest making this a major version bump, to 7.0.0.

makermelissa (approved):
Thanks, I like the idea of this being an external library.

PR #560, opened 2022-03-01
usb_hid module

fivetide:
This PR is going to provide a usb_hid adapter using usb gadget. It is not yet complete, but i would like to collect early input on it.

- [x]  Add device class
- [x] usb_hid.disable() and usb_hid.enable()
- [x] can send reports
- [x] can receive reports (get_last_received_report)
- [x] sane descriptor for Device.KEYBOARD
- [x] sane descriptor for Device.MOUSE
- [x] sane descriptor for Devices.CONSUMER_CONTROL
- [x] boot device...?

### DIFF
--- a/src/usb_hid.py
+++ b/src/usb_hid.py
@@ -1,0 +1,246 @@
+from typing import Sequence
+from pathlib import Path
+import atexit
+
+# https://www.kernel.org/doc/Documentation/usb/gadget_configfs.txt
+
+gadget_root = '/sys/kernel/config/usb_gadget/adafruit-blinka'
+
+
+class Device():
+    def __init__(self, *, descriptor: bytes, usage_page: int, usage: int, report_ids: Sequence[int],
+                 in_report_lengths: Sequence[int], out_report_lengths: Sequence[int]) -> None:
+        self.out_report_lengths = out_report_lengths
+        self.in_report_lengths = in_report_lengths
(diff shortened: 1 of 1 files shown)


dhalbert:
The report descriptors we use in CircuitPython are here:
https://github.com/adafruit/circuitpython/blob/main/shared-module/usb_hid/Device.c

fivetide:
So i was able to verify this works on a raspi zero wh with kmk firmware. Some things remain untested, like using multiple reports on one single device. 

What i did not grasp: Is the user supporsed to call usb_hid.enable manually, or would this have to be automated?

makermelissa:
Based on https://learn.adafruit.com/customizing-usb-devices-in-circuitpython/hid-devices, I would say the user is supposed to call it manually.

makermelissa:
The only thing is in circuitpython, it uses a boot.py file and we really don't have the equivalent in Blinka, so I think having it in the it manually enabled user script makes the most sense because many boards probably wouldn't support this.

makermelissa (approved):
Thank you for adding this. I know there are several folks who will appreciate this.

makermelissa:
> what are prerequirements for this module on raspberry pi zero? Is there any example code?

I'm. not sure since I haven't run it. I think adding an example would be a good idea and this should be a new issue so it doesn't fall between the cracks.

Issue #575, opened 2022-04-13
rp2040: Missing I2C ports


Gadgetoid:
The `i2cPorts` list for rp2040 is incomplete.

See: https://github.com/adafruit/Adafruit_Blinka/blob/c4d26d1a2f3d037a20aa5ae55936baa265258edb/src/adafruit_blinka/microcontroller/rp2040/pin.py#L91-L124

Versus available combinations detailed on: https://pico.pinout.xyz/

Is this intentional? It looks like the ports use permutations which means a complete list would be rather a lot longer.

I had a user attempting to use I2C on GP21 / GP20 which is a valid combination for i2c0 and this failed with a very long-winded error:

```
ValueError: No Hardware I2C on (scl,sda)=(board.GP21, board.GP20)
Valid I2C ports: ((0, board.GP1, board.GP0), (0, board.GP1, board.GP4), (0, board.GP1, board.GP8), (0, board.GP1, board.GP12), (0, board.GP5, board.GP0), (0, board.GP5, board.GP4), (0, board.GP5, board.GP8), (0, board.GP5, board.GP12), (0, board.GP9, board.GP0), (0, board.GP9, board.GP4), (0, board.GP9, board.GP8), (0, board.GP9, board.GP12), (0, board.GP13, board.GP0), (0, board.GP13, board.GP4), (0, board.GP13, board.GP8), (0, board.GP13, board.GP12), (1, board.GP3, board.GP2), (1, board.GP3, board.GP6), (1, board.GP3, board.GP10), (1, board.GP3, board.GP14), (1, board.GP7, board.GP2), (1, board.GP7, board.GP6), (1, board.GP7, board.GP10), (1, board.GP7, board.GP14), (1, board.GP11, board.GP2), (1, board.GP11, board.GP6), (1, board.GP11, board.GP10), (1, board.GP11, board.GP14), (1, board.GP15, board.GP2), (1, board.GP15, board.GP6), (1, board.GP15, board.GP10), (1, board.GP15, board.GP14))
```

An exhaustive list of port permutations might get unwieldy. Would it be appropriate to - in the case of the Pico/RP2040 - replace this with the logic used in the SDK, MicroPython and CircuitPython C/C++ code.

IE: since i2c is available on (almost) all pins with a common pattern the `SCL % 2 == 1 and SDA % 2 == 0` check should suffice.

[...]

PR #576, opened 2022-04-14
RP2040: Add missing i2c pin permutations.

Gadgetoid:
Add the missing i2c pin permutations as discussed in #575.


makermelissa (approved):
Looks good to me.

Issue #573, opened 2022-03-29
MC2221A import board RuntimeError: already open


nickGermi:
Hello,

When running following python code
`import os; os.environ['BLINKA_MCP2221'] = '1'`
`import os; os.environ['BLINKA_MCP2221_RESET_DELAY'] = '0.5'`
`import board`

I get following error:

> Traceback (most recent call last):
>   File "c:\Users\Nick\Desktop\mc2221a.py", line 8, in <module>
>     import board
>   File "C:\Python39\lib\site-packages\board.py", line 182, in <module>
>     from adafruit_blinka.board.microchip_mcp2221 import *
>   File "C:\Python39\lib\site-packages\adafruit_blinka\board\microchip_mcp2221.py", line 5, in <module>
>     from adafruit_blinka.microcontroller.mcp2221 import pin
>   File "C:\Python39\lib\site-packages\adafruit_blinka\microcontroller\mcp2221\pin.py", line 5, in <module>
>     from .mcp2221 import mcp2221
>   File "C:\Python39\lib\site-packages\adafruit_blinka\microcontroller\mcp2221\mcp2221.py", line 389, in <module>
>     mcp2221 = MCP2221()
>   File "C:\Python39\lib\site-packages\adafruit_blinka\microcontroller\mcp2221\mcp2221.py", line 58, in __init__
>     self._reset()
>   File "C:\Python39\lib\site-packages\adafruit_blinka\microcontroller\mcp2221\mcp2221.py", line 128, in _reset
>     self._hid.open(MCP2221.VID, MCP2221.PID)
>   File "hid.pyx", line 111, in hid.device.open
> RuntimeError: already open

If I just set environment variable and open a Python shell and just type "import board" I get same exact error as above.

If I try to use PyMCP2221A module to communicate with my MCP221A device instead, it seems to be working fine. "MCP2221 I2C/SMBus Terminal" utility also seems to be working fine (I can read data via I2C protocol)

I'm using latest version of adafruit-blinka (7.1.1) & hidapi (0.11.2). I was using following guide to set everything up:
https://learn.adafruit.com/circuitpython-libraries-on-any-computer-with-mcp2221/windows
My environment is Windows 11

nickGermi:
So after commenting out following lines:

`lib\site-packages\adafruit_blinka\microcontroller\mcp2221\mcp2221.py`
```
61        # if MCP2221_RESET_DELAY >= 0:
62        #     self._reset()
```

[...]

PR #577, opened 2022-04-14
Close HID connnection to MCP2221 when resetting

caternuson:
Fix for #573.

Pretty simple, just need to explicitly close when resetting. Tested locally to recreate original issue and verify this patch worked to fix. Even blinked an LED to make sure.


caternuson:
Oh, wait a sec. There's U2IF also. Will check that and push some more updates.

caternuson:
Yep. Same thing happens with U2IF. Also get the "already open" error. Used a Feather RP2040 to verify. Closing fixes there as well. Blinked #13 led to verify.

ladyada (approved):
nice find

caternuson:
thanks. wasn't me:
https://github.com/adafruit/Adafruit_Blinka/issues/573#issuecomment-1098404088

PR #579, opened 2022-04-29
Allow Blinka to run on non-embedded environments

shulltronics:
This PR allows blinka to run on non-embedded systems, i.e. GENERIC_LINUX or Windows Environments. The motivation was to allow the use of the displayio library on non-circuitpython boards. (See https://github.com/adafruit/Adafruit_Blinka_Displayio/pull/63 for a discussion about this use-case). This allows the library to be used to create general-purpose desktop GUIs, as well as to simulate a hardware display so that embedded GUIs can be developed and tested without physical hardware.

Instead of failing with a `NotImplementedError` if the platform's `chip_id` is `None` or `'GENERIC_X86'`, I print a warning message. In order to satisfy the `Pin` import in `pin.py`, I imported the class from `generic_micropython`. This might not be the best thing to do here, since the platform is not necessarily micropython -- I'm open to suggestions about how to make this more semantically accurate.

### DIFF
--- a/src/microcontroller/__init__.py
+++ b/src/microcontroller/__init__.py
@@ -119,7 +119,7 @@ elif chip_id == ap_chip.RP2040_U2IF:
 elif chip_id == ap_chip.GENERIC_X86:
     print("WARNING: GENERIC_X86 is not fully supported. Some features may not work.")
 elif chip_id == None:
-    print("WARNING [__init__.py]: chip_id is None!")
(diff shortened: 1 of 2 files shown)


FoamyGuy (approved):
Changes look good to me. Tested successfully with https://github.com/FoamyGuy/Blinka_Displayio_PyGameDisplay on Linux OS.

makermelissa (approved):
Thanks. Looks good.

PR #581, opened 2022-05-02
Require adafruit-circuitpython-typing on install

### LINKED ISSUE #580: Blinka does not install adafruit-circuitpython-typing
Release 7.0.0 sounds like it made it so `adafruit-circuitpython-typing` but it wasn't added to `setup.py` or `requirements.txt`.  I think a lot of the libraries (at least the ones I've seen/modified) explicitly call out for that module in their own `setup.py` and `requirements.txt`, so it hasn't been a huge issue.  Should it be added here and automatically installed with Blinka, or only installed once a package is installed that makes use of those types?

tekktrik:
Fixes #580 by adding `adafruit-circuitpython-typing` to `setup.py` and `requirements.txt`.  I don't think a specific version is needed, and the few libraries that require features that weren't in base (really anything other than `ReadableBuffer` and `WriteableBuffer`) and had them added to that library will have to call those out and pin to minimum versions anyway.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ sysv_ipc>=1.1.0; sys_platform == 'linux' and platform_machine!='mips'
 pyftdi>=0.40.0
 binho-host-adapter>=0.1.6
 numpy>=1.21.5
+adafruit-circuitpython-typing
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Thanks

PR #586, opened 2022-05-26
Added usb_hid to py_modules in setup.py

### LINKED ISSUE #585: ModuleNotFoundError: No module named 'usb_hid'
I think this may be because `usb_hid` is not included in the setup script here, so it's not being added to the package https://github.com/adafruit/Adafruit_Blinka/blob/main/setup.py#L58

makermelissa:
Fixes #585.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ setup(
         "pulseio",
         "pwmio",
         "rainbowio",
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Thanks!

PR #590, opened 2022-06-03
Move (unimplemented) OneWire from busio to onewireio

dhalbert:
We have moved `OneWire` to its own module in CircuitPython, so we can easily turn it on and off

In CircuitPython 6.x.x, `OneWire` was available  as `busio.OneWire`.
In CircuitPython 7.x.x, `OneWire` is available as `busio.OneWire` _and_ as `onewireio.OneWire`.
In CircuitPython 8.x.x, `OneWire` will be available  only as `onewireio.OneWire`.

Have Blinka catch up to `onewireio.OneWire`. Note that `OneWire` is still unimplemented in Blinka

@makermelissa Do you want this to be a major version change? It's an API change, but for a piece of the API that's unimplemented. But integers are free.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -40,6 +40,9 @@
 .. automodule:: neopixel_write
   :members:
 
+.. automodule:: onewireio
+  :members:
+
 .. automodule:: pulseio
   :members:
 
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ setup(
         "keypad",
         "micropython",
         "neopixel_write",
+        "onewireio",
         "pulseio",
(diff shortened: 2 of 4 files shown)


makermelissa:
Yes, let's do a major version change in case (however unlikely) somebody was making use of OneWire in busio for whatever reason.

makermelissa (approved):
Thanks

ladyada:
fyi we don't/can't have onewire support in blinka cause there's no low level 1wire kernel support - so like for ds18b20 they have to use the specific kernel sysfs driver/

PR #591, opened 2022-06-06
Fix Raspberry Pi SPI bus 1 not working

tinue:
The Raspberry Pi SPI bus 1 can't be used with the Blinka library. I suspect a simply typo.
With the supplied patch it works as expected. See also https://github.com/tinue/apa102-pi/issues/52


makermelissa (approved):
Yep, it looks like the typo happened in https://github.com/adafruit/Adafruit_Blinka/commit/bb9dd9909c672798b9d69c8e3689b2e443c8d936

Thanks for fixing.

PR #597, opened 2022-07-22
Add Banana Pi M5 support.

djhedges:
This pull request is dependent on https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/236 which populates the board and chip values.  Testing was done with the adafruit_fxos8700 library and fxos8700 accelerometer.


makermelissa:
> I should point out I blindly copied board/hardkernel/odroidc4.py's pinout for bpim5.py. They're both using the s905x3 chipset so I assumed that's fine. I didn't fully understand the documentation at https://learn.adafruit.com/adding-a-single-board-computer-to-blinka?view=all

That's fine as a starting point. Did you test that the physical pins maps to the pins you think they do like in https://learn.adafruit.com/adding-a-single-board-computer-to-blinka?view=all#testing-a-physical-gpio-pin-3059113?

I went ahead and added the added in your Platform Release code and restarted the checks, so hopefully they'll pass fine now.

djhedges:
> > I should point out I blindly copied board/hardkernel/odroidc4.py's pinout for bpim5.py. They're both using the s905x3 chipset so I assumed that's fine. I didn't fully understand the documentation at https://learn.adafruit.com/adding-a-single-board-computer-to-blinka?view=all
> 
> That's fine as a starting point. Did you test that the physical pins maps to the pins you think they do like in https://learn.adafruit.com/adding-a-single-board-computer-to-blinka?view=all#testing-a-physical-gpio-pin-3059113?
> 
> I went ahead and added the added in your Platform Release code and restarted the checks, so hopefully they'll pass fine now.

I have not yet.  I'll go throw the pins later today and double check them.

makermelissa:
Awesome, thanks. Let me know if you have any other questions or see parts of the guide that are unclear and I'll see if I can improve them. Thanks again.

djhedges:
PTAL

The hardware side of things I'm really fuzzy on.  So to test the pins I wrote a small script like so.  The `da02160` commit has comments with "#Verified" which are the pins I was able to blink an LED with.  The other pins had a dual purpose? and would throw an error or not blink the led.
```
#!/usr/bin/python3

import sys
import time
import board
import digitalio

pin_num = sys.argv[1]
pin = f'P{pin_num}'
print(f'Testing {pin}')
led = digitalio.DigitalInOut(getattr(board, pin))
led.direction = digitalio.Direction.OUTPUT
led.value = True
time.sleep(2)
led.value = False
```

I did use the adafruit_dotstar.DotStar library to verify the board.SCLK & board.MOSI pins.

I used this documentation to update several of the pins which corrected a few issues.  Such as there only be 40 pins.  I also went ahead and renamed the pins from D# > P# to correspond with that documentation.
https://wiki.banana-pi.org/Banana_Pi_BPI-M5#BPI-M5_40PIN_GPIO_.28CON2.29

One thing I found very strange is two pins have the GPIOAO_7 according to that wiki.  I wonder if that's a typo.

djhedges:
Yep that wiki is incorrect for pin 37.  It should be GPIOAO_9 which now lights up the LED.  This kernel patch reflects that.
https://patchwork.kernel.org/project/linux-arm-kernel/patch/20210429170404.3616111-4-narmstrong@baylibre.com/

makermelissa (approved):
Looks good. Since you used an existing microcontroller, the busio stuff should just work . Thanks.

PR #606, opened 2022-09-05
usb_hid: open in binary does not take an encoding

Neradoc:
As per python docs:
> For reading and writing raw bytes use binary mode and leave encoding unspecified

```py
>>> open("test", "rb+", encoding="utf-8")
ValueError: binary mode doesn't take an encoding argument
>>> 
```

### DIFF
--- a/src/usb_hid.py
+++ b/src/usb_hid.py
@@ -60,7 +60,7 @@ class Device:
         """
         report_id = report_id or self.report_ids[0]
         device_path = self.get_device_path(report_id)
-        with open(device_path, "rb+", encoding="utf-8") as fd:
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Ah, this makes sense. I think it may have been added because of PyLint recommending it at some point.

Med-JH:
thank you @makermelissa

makermelissa:
> thank you @makermelissa

It was @Neradoc who made the PR. :) I just reviewed it.

Med-JH:
> > thank you @makermelissa
> 
> It was @Neradoc who made the PR. :) I just reviewed it.

then tank you @Neradoc aswell

PR #608, opened 2022-09-06
Add typing information to micropython

jepler:
This requires converting it into a package so that it can have a "py.typed" file.

After this, a package using micropython.const can be type checked

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,8 @@ setup(
         "adafruit_blinka.microcontroller.bcm283x.pulseio": [
             "libgpiod_pulsein",
             "libgpiod_pulsein64",
-        ]
+        ],
+        "micropython": ["py.typed"],
     },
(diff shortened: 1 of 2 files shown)


makermelissa:
Thanks. It looks like this change should be fine. Have you tested on MicroPython and Linux? Blinka has been super touchy about adding typing info in the past because it's supposed to work in such a wide variety of environments. 

In fact, in the future, I think splitting it up a bit more internally I think would be helpful with fixing this touchiness. I'll make another issue regarding this in more detail.

jepler:
No, I did not test anything.

An alternative is to provide micropython.pyi, I could PR that approach instead.

jepler:
Let's just go the #610 route.

makermelissa:
Sounds good. Thanks.

PR #610, opened 2022-09-06
Provide stubs for micropython module in a more clearly compatible way

jepler:
This allows use of micropython.const to typecheck without making any changes to the .py files themselves. It's an alternative to #608


tekktrik (commented):
Just chiming in on the typing - looks good.  Stylistically the returns can be replaced with `...` in `.pyi` files, but it shouldn't be a problem as is (I think).

tekktrik (approved):
Looks good to me!

PR #616, opened 2022-09-20
Fix build missing in release CI

tekktrik:
This should fix the problem with `build` missing.  Tested locally and worked.


makermelissa (approved):
Thanks

makermelissa:
Yep, awesome.

Issue #617, opened 2022-09-20
Libre AML-S905X-CC (LePotato) 


Snakn1:
Can you please add the  Libre AML-S905X-CC (LePotato) the pin info can be found at [https://github.com/libre-computer-project/libretech-wiring-tool/blob/master/libre-computer/aml-s905x-cc/gpio.map](https://github.com/libre-computer-project/libretech-wiring-tool/blob/master/libre-computer/aml-s905x-cc/gpio.map)

or 
[https://docs.google.com/spreadsheets/d/1U3z0Gb8HUEfCIMkvqzmhMpJfzRqjPXq7mFLC-hvbKlE/edit?usp=sharing](https://docs.google.com/spreadsheets/d/1U3z0Gb8HUEfCIMkvqzmhMpJfzRqjPXq7mFLC-hvbKlE/edit?usp=sharing)

makermelissa:
If you'd like to try, here are the guides that go over how to add a board to Blinka:
https://learn.adafruit.com/adding-a-single-board-computer-to-platformdetect-for-blinka
https://learn.adafruit.com/adding-a-single-board-computer-to-blinka

Snakn1:
I am sorry I am not sure how to reply so if I need to reply on GitHub then
let me know.

I can not detect the cpu or the board type per your directions
using detect.py.

I am running "ubuntu-22.04.1-preinstalled-desktop-arm64+aml-s905x-cc"


Given the information I have, what is the next step?


sudo python3 detect.py

Chip id:  None
Board id:  None

Is this a DragonBoard 410c? False
Is this a Pi 3B+? False
Is this a Pi 4B? False
Is this a 40-pin Raspberry Pi? False
Is this a Raspberry Pi Compute Module? False
Is this a BeagleBone Board? False
Is this a Giant Board? False
Is this a Coral Dev Board? False
Is this a Coral Dev Board Mini? False
Is this a MaaXBoard? False
Is this a MaaXBoard Mini? False
Is this a SiFive Unleashed?  False
Is this a PYNQ Board? False
Is this a Rock Pi board? False
Is this a NanoPi board? False
Is this a Khadas VIM3 board? False
Is this a Clockwork Pi board? False
Is this an embedded Linux system? False
Is this a generic Linux PC? False
Is this a UDOO Bolt? False
Is this an ASUS Tinker Board? False
Is this an STM32MP1 Board? False
Is this an OS environment variable special case? False


more /proc/cpuinfo

[...]

PR #620, opened 2022-10-15
Add support for aml-s905x-cc-v1

bmol118:
Add support for aml-s905x-cc-v1.

I tested all of the pin mappings & verified they worked. 
Currently this is failing because it is not building with the [corresponding change](https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/252) in Adafruit_Python_PlatformDetect. 

There is currently an issue open for this: [issue 617](https://github.com/adafruit/Adafruit_Blinka/issues/617)


makermelissa (approved):
Looks good to me. Thanks.

PR #627, opened 2022-11-07
Fix digitalio maintenance by removing non-special case boards

makermelissa:
Digitalio was supposed to be mostly chips from the platform detect because we're importing the chip info here. This fixes that.


kattni (approved):
Looks good to me!

PR #632, opened 2022-11-26
Added support to SeeedStudio Odyssey X86J4105 SBC

ecarozzo:
Added support for SBC "Odyssey X86J4105" from SeeedStudio, SPI not supported because the board use Spidev 1 on header pins and only Spidev 0 is supported from the code.


ecarozzo:
Hi,
i need review of [my pull request](https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/263) because this check fail due to lack of import from PlatformDetect.

Thanks

makermelissa (approved):
Thanks for adding. Looks great!

PR #635, opened 2022-11-30
 Add support for Radxa rock 5b

happyme531:
RK3588 is a recent & very powerful SoC with rich peripherals. 
This pull request adds the RK3588 Chip and Rock 5B support.
To prevent confusion since many peripherals pin can be muxed to different GPIOs, the periphal pin name is picked directly from the datasheet, Mx means "Mode x"


tannewt:
Typically we wouldn't have pin names for the Mx versions. Instead, you just pass the constructor the pins you want and it'll determine the correct muxing or throw an error.

happyme531:
> Typically we wouldn't have pin names for the Mx versions. Instead, you just pass the constructor the pins you want and it'll determine the correct muxing or throw an error.

But then some peripheral pin, eg.UART2_TX will be corresponding to two pins, and there are no default one. So what to do next?

makermelissa:
> > Typically we wouldn't have pin names for the Mx versions. Instead, you just pass the constructor the pins you want and it'll determine the correct muxing or throw an error.
> 
> But then some peripheral pin, eg.UART2_TX will be corresponding to two pins, and there are no default one. So what to do next?

I think what Scott meant is that Blinka doesn't usually handle any muxing itself. Often time the device tree that's loaded handles that or possibly even at the kernel level. It's possible to write a specific Pin implementation to mux the pin correctly, but I don't believe there are currently any boards that do this.

tannewt:
Blinka is meant to be like CircuitPython and in CircuitPython, it is the job of the UART implementation to set the mux for the given mcu pin. Internal peripheral pins aren't referenced in Blinka or CP code.

makermelissa (approved):

PR #641, opened 2022-12-14
Fix board name after update to latest bios firmware

ecarozzo:


makermelissa (approved):
Looks good. I forgot about this needing to be changed here with the PlatformDetect PR.

PR #646, opened 2023-01-17
Added missing if query for Siemens SIMATIC IOT2050 Basic Version -> updated board.py

schnurma:
The "if query" for the SIMATIC IOT2050 Basic was missing.
It could be tested now with the new version of the SIMATIC IOT2050 Basic (FS:04 12.2022)

thanks
best regards
Martin Schnur


makermelissa (approved):
Looks good. Thanks.

PR #647, opened 2023-01-19
Minimal support for generic linux pc

bablokb:
This pull requests adds minimal support for boards with id 'GENERIC_LINUX_PC'. With this patch, an `import board` no longer fails and `board.board_id` is something to work with.

I am using this together with @FoamyGuy s <https://github.com/FoamyGuy/Blinka_Displayio_PyGameDisplay> package to run CircuitPython-code on a normal linux-box (mainly to create screenshots). With this patch, `import board` does not fail anymore.

### DIFF
--- a/src/board.py
+++ b/src/board.py
@@ -11,7 +11,7 @@ See `CircuitPython:board` in CircuitPython for more details.
 """
 
 
-__version__ = "0.0.0+auto.0"
+__version__ = "8.12.0"
 __repo__ = "https://github.com/adafruit/Adafruit_Blinka.git"
 __blinka__ = True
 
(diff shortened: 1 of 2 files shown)


src/board.py
```diff
@@ -11,7 +11,7 @@
 """
 
 
-__version__ = "0.0.0+auto.0"
+__version__ = "8.12.0"
```
FoamyGuy, line 14: I think these versions are supposed to stay on the 0s / auto. It gets filled in by something further down the chain I believe.

tekktrik: Yup!  The CI finds that string and replaces it so it can be inside the wheels/scource distributions.  It gets replaced with whatever the GitHub Release tag is.

FoamyGuy (approved):
This looks good to me, and I am in favor of adding it even though it's not implementing any more complex functionalities. 

Being able to `import board` on generic linux without an exception being raided is nice because it helps increase the portability of code, in particular displayio tests and scripts that can run on microcontrollers and raspi / desktop using the pygame display for output.

I think it would good to get input from @makermelissa as well if possible.

I did successfully test this change on my PC.

PR #651, opened 2023-02-08
Add compatibility with Radxa's Rock Pi 4 C Plus

wgroeneveld:
Hi,

I have a recent Radxa Rock Pi 4, but it's the C+ model, which, although compatible/recognizable by [Adafruit_Python_PlatformDetect](https://github.com/wgroeneveld/Adafruit_Python_PlatformDetect), is not currently compatible with Adafruit_Blinka.

I've added a GPIO mapping according to https://wiki.radxa.com/Rock4/hardware/gpio (press EXPAND to see the 4C+ GPIO mapping). I've tested I2C7, which works, but weirdly, `mraa-gpio list` lists slightly different fields: GPIO4_A7 is listed on pin D40, but according to Radxa's page, pin D40 should be GPIO1_C4. 

I'm aware of GPIO problems on 4C+ models (see https://forum.radxa.com/t/pi4cplus-gpio-not-working/14090/8 ) so not sure who to trust: the official docs or the `mraa-gpio` output. For now, I adhered to the first. 

Thanks for considering to merge this!


makermelissa:
Hi, I noticed you changed D17 to D18 on the Rock Pi 4 board.  I'm just curious if it was incorrect or for another reason? Thanks.

Also, it looks like you're super close to getting it to pass the checks.

makermelissa:
Never mind, I see. Yeah, it was a voltage pin. All good. :)

wgroeneveld:
> Never mind, I see. Yeah, it was a voltage pin. All good. :)

Yeah that was incorrect so I fixed that one!

wgroeneveld:
I can't figure out why the build is failing, is there anything I can do to help get this through?

makermelissa (approved):
Looks good. Thank you.

RadxaYuntian:
Hi @wgroeneveld, the pinout you are referencing is for ROCK 4C+ v1.0/1.1, and those are pre-production versions. The mass production began with v1.2, which fixed the pinout to be compatible with the rest of ROCK 4 family.

I think in your [rockpi4cplus.py](https://github.com/adafruit/Adafruit_Blinka/pull/651/files#diff-df8ac62c5cd5b1665ab0c90099b0540cdcbae540885b4290282022276531497d) you are using this pre-production pinout, which will not work correctly on production versions.

makermelissa:
@RadxaYuntian, thanks for the note. 

@wgroeneveld it looks like we should add detection for preprod and regular versions and load the pinout appropriately [here](https://github.com/adafruit/Adafruit_Python_PlatformDetect/blob/main/adafruit_platformdetect/board.py)#L526-L527 if you'd like to give it a shot. I'd just see what info you can pull by trying out the functions and see what you can get from the device tree. I'd probably look for something like 1.0 or 1.1 in the data returned. This way we can ensure that both versions work.

PR #662, opened 2023-03-09
Add support for missing rp2040 SPI0 RX (GP20) pin

FlantasticDan:
I noticed the SPI0 RX capability of the GP20 pin was throwing an error in MicroPython via Blinka but not in native CircuitPython on a Pico W.  I'm specifically using `busio.SPI(board.GP18, MOSI=board.GP19, MISO=board.GP20)` and have confirmed this PR fixes that pin combination.  I used the [Pico W Pinout](https://pico.pinout.xyz/) to infer the remaining SPI0 combinations.


makermelissa (approved):
Thank you.

PR #658, opened 2023-03-03
Fix Rock 4C Plus pin mapping and board file

MathijsNL:
This will fix issue #656.

I tested all pins as input and output, they all work except for the not so well documented D26. The information on the wiki matches the actual layout.
[Radxa gpio](https://wiki.radxa.com/Rockpi4/hardware/gpio)

@RadxaYuntian do you know where we can find more info about pin# 26 (ADC IN)?

PR #664, opened 2023-03-15
ROCK_PI_4 improvements

### LINKED ISSUE #659: Pin numbers on Rock4 not matching with actual device
When working on the Rock 4C+ I noticed that the pinout for the Rock 4 was not matching the pinout on the [Radxa wiki](https://wiki.radxa.com/Rock4/hardware/gpio).

I just tested the Rock4 B+ and it seems that the pin mapping is the same. All pins were tested as output with a LED.

The quickest fix is to just copy over the content from the Rock 4C+ pin and board file to the Rock 4B+ file. I think it would be better though to just have the board.py point these 2 boards to the same files:

``` 
elif board_id == ap_board.ROCK_PI_4:
    from adafruit_blinka.board.radxa.rockpi4 import *

elif board_id == ap_board.ROCK_PI_4_C_PLUS:
    from adafruit_blinka.board.radxa.rockpi4 import * # <-- rockpi4 instead of rockpi4cplus
```

It seems the full Rock4 family is pin compatible (except for the 4C+ pre-prod version) so this would probably work. I don't have the 4C+ version that has the old layout, so if someone really needs to make that board compatible I think it can always be added again as ROCK_PI_4_C_PLUS_PREPROD if someone is willing to take the effort of doing so. 

If someone (@makermelissa  ?) could share their opinion about the best approach here I will finish the work and create a PR.

MathijsNL:
This PR will both fix the ROCK_PI_4 pinout to match the [wiki](https://wiki.radxa.com/Rockpi4/hardware/gpio) and it will make the Rock 4 family work on a single board and pin file. Files were already tested on 4C+ and 4B+ in PR #658 . I tested a few GPIO pins after making the changes for this PR to make sure everything works as intended and it seems to be ok.

This PR will fix #659


MathijsNL:
It seems like the orange pi 4 is also using the RK3399 chip. This PR won't work as long as the conflict with Orangepi 4 is not resolved.

In my opinion it is important to have a proper detection for Orange pi 4, because even though they use the RK3399 chip has not much to do with their pin layout.

Since I don't have an Orange pi 4 around there is not much I can do about it. Anyone has a solution for this?

MathijsNL:
It seems the Orange Pi board detection is working properly, I renamed the original rk3399 to rk3399_orangepi4 to solve the issue. The PR as it is right now should fix the complete Rock 4 family without breaking the Orange PI 4.

MathijsNL:
Maybe a small oversight, I think the board_id has to be added as a check to some of the files as well as to prevent the Orange Pi 4 from breaking. I will have a look at it later this week.

makermelissa (changes requested):
Hi, thanks for this. I would prefer to have the chip files combined if possible. When you talk about a conflict between pins on the Orange Pi 4 and Rock4, which pins are you referring to? I noticed that one board uses some pins and another uses other pins, but didn't notice pins with the same name that point to different locations.

If you're referring to the I2C or SPI pins, would it be possible to include everything and then adjust the board file refer to the correct pins?

makermelissa:
The reason why we have separate board and microcontroller files is because the microcontroller file maps things like when you are accessing GPIO4_C6, you are actually accessing sysfs pin 150. These values should be constant with the same microcontroller even on different boards. The board file maps what pins are used on the 40-pin header to the microcontroller pin. So for instance referring to the pinouts you posted, pin 13 on one board accesses GPIO4_C6 and a different board it has GPIO1_A3 for pin 13. Both boards make use of GPIO4_C6 in different spots and it maps to 150 in both cases.

While the Orange Pi 4 board only has 1 PWM pin (Pin 7), the microcontroller itself is capable of having 2 PWM pins. The second is just not used.

MathijsNL:
So if I understand correctly it shouldn't matter if the GPIO numbers are different, it is the chip that matters. So for the RK3399 we should have a mapping to all the pins, even if they are different for different boards, and then use the board file to define which ones actually have to be used?

makermelissa:
Exactly. :)

MathijsNL:
Things were getting a bit messy so I force pushed as to not have so many back and forth changes.

I think the chip files are merged correctly now, and there is no 4cplus board anymore and the orange pi should still work.

Thanks for your input @makermelissa, in the end working with 1 chip file was actually a simpler solution.

makermelissa (approved):
This looks great. Thank you.

PR #665, opened 2023-03-24
Added minimal beaglebone blue support

jath03:
Hi, I needed to use I2C on the BeagleBone Blue, so I added support for that in the library.  This pr only includes pins that I have tested personally, but someone with more knowledge about the blue might be able to fill in the blanks.  According to [this](https://github.com/adafruit/adafruit-beaglebone-io-python/issues/313#issuecomment-513904700), it might be pretty similar to the black, but I don't know anything for sure.  Would the maintainers want a partially supported board like this included in the official library or kept as a fork?


makermelissa (approved):
Thanks for adding. Hopefully somebody who has one of these boards will add the GPIO pins later.

PR #670, opened 2023-04-07
Add second I2C to coral_dev_board.py

stevengilbertoksi:
I needed to use the second I2C port (I2C3) and so I added this locally, and it works.


tannewt (approved):
Thank you!

PR #671, opened 2023-04-11
Fixing swapped SPI0/SPI1

BlitzCityDIY:
Fixing swapped SPI0/SPI1 for EPD Feather


tekktrik (approved):
Didn't test but if the swap is needed then it looks food to me!

tekktrik:
It also looks *good* to me

PR #672, opened 2023-04-11
Add option for I2C frequency

joshi-bharat:
Added option to pass I2C frequency to I2C constructor

### DIFF
--- a/src/board.py
+++ b/src/board.py
@@ -356,11 +356,11 @@ else:
 
 if "SCL" in locals() and "SDA" in locals():
 
-    def I2C():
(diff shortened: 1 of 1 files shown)


makermelissa:
Thanks for the PR. However, this was purposefully left out for compatibility with CircuitPython (see https://docs.circuitpython.org/en/latest/shared-bindings/board/index.html) which is explicitly intended to only use the defaults. You should be able to use the following to set it to 100000:

```
import board
import busio

i2c = busio.I2C(board.SCL, board.SDA, frequency=100000)
```

PR #673, opened 2023-04-11
Adding support for CAN and RFM Feathers

BlitzCityDIY:
Adding support for CAN and RFM Feathers to use with U2IF


makermelissa (approved):
Looks good to me. You can either merge if you're happy with it or request Carter to review as well.

Issue #656, opened 2023-02-26
Pin numbers for Rock 4C Plus not available


MathijsNL:
I am using the library on a Rock 4C Plus. I am trying to use GPIO 73 and 74 (pin 31 and 29) 

Blinka version = '8.15.2'
Python version = 3.8.10
OS = Ubuntu 20.04 arm64

```
>>> board.board_id
'ROCK_PI_4C+'
```

Pin assignment:
```
outPin = board.B1
AttributeError: module 'board' has no attribute 'B1'
```

As I understand the mapping is available here:
https://github.com/adafruit/Adafruit_Blinka/blob/main/src/adafruit_blinka/microcontroller/rockchip/rk3399/pin.py#L24-L25

My assumption is that I need to use pin board.B1 and board.B2, but they don't show up. All available pins are shown below:

```
board.ADC_IN0   board.D27       board.D8        board.SDA7
board.CS        board.D28       board.I2C(      board.SPI(
board.D10       board.D29       board.MISO      board.UART2_RX
board.D11       board.D3        board.MOSI      board.UART2_TX
board.D12       board.D31       board.PWM0      board.UART4_RX
board.D13       board.D32       board.SCK       board.UART4_TX
board.D15       board.D33       board.SCL       board.UART_RX
board.D16       board.D35       board.SCL2      board.UART_TX
board.D18       board.D36       board.SCL6      board.ap_board
board.D19       board.D37       board.SCL7      board.board_id
board.D21       board.D38       board.SCLK      board.detector
board.D22       board.D40       board.SDA       board.pin
board.D23       board.D5        board.SDA2      board.sys
board.D24       board.D7        board.SDA6

board.pin
<module 'adafruit_blinka.microcontroller.rockchip.rk3399_t.pin' from '/home/rock/.local/lib/python3.8/site-packages/adafruit_blinka/microcontroller/rockchip/rk3399_t/pin.py'
```

@wgroeneveld Tagging you if you don't mind. Do you happen to know what the problem is?

makermelissa:
It's likely related to this: https://github.com/adafruit/Adafruit_Blinka/pull/651#issuecomment-1439630083, meaning the pinout is was added incorrectly for your board. I'm guessing your board is a production board (the one that was used to figure out the pinout wasn't), so maybe with your help, we could get it working (hopefully with both versions). If we can't get it working with both versions, I'd prefer the production board be the one that works.

MathijsNL:
[...]

PR #674, opened 2023-04-16
Fix Incorrect SPI Pin Assignments for Rockpi4

MathijsNL:
This will solve #656 (I think it should have been closed already anyway, my bad)

The documentation from Radxa shows that the mapping was indeed wrong: https://wiki.radxa.com/Rock4/hardware/gpio


makermelissa (approved):
Thanks, looks good to me.

PR #691, opened 2023-07-22
added support for i2c1-m4 overlay in orange pi 5

ErikApption:
The rk3588 I2C IDs do not seem to match the orange pi 5 ones - I added support for i2c1-m4 and tested it successfully with a BMP390. I changed default to i2c1-m4


makermelissa (approved):
Thanks.

PR #703, opened 2023-09-07
Add issue templates and update pre-commit config

### LINKED ISSUE #701: Update unrecognized board error to suggest filing an issue
I would like to keep new board requests added to Blinka (unless detection only is needed).

makermelissa:
This informs to submit a new board request if their board isn't supported, add issue templates and updates the versions in the pre-commit config. Fixes #701.


makermelissa:
Due to fixing a constant name, this should be merged around the same time as https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/314.

PR #708, opened 2023-09-12
Display a message if user tries to set freq on linux

### LINKED ISSUE #682: Why can't I2C baudrate/frequency be set in Python?
For all microcontrollers supported by this project, there are lines commented out that state that the I2C frequency cannot be set in Python and is therefore ignored. because these print statements are commented out, the end user (me) has no idea why the API parameter isn't working.
https://github.com/adafruit/Adafruit_Blinka/blob/main/src/adafruit_blinka/microcontroller/generic_linux/i2c.py#L23-L24

I'm using a Raspberry Pi Zero W with the Adafruit motor control bonnet to control a stepper motor which is turning slower than I like. The Adafruit MotorKit documentation suggests updating the frequency to get more speed.
https://docs.circuitpython.org/projects/motorkit/en/latest/api.html#adafruit_motorkit.MotorKit

I've already changed the `config.txt` setting to update the I2C speed from 100khz to 400khz, but my understanding is that the code also needs to use it. I've confirmed that it's now 400khz.
https://learn.adafruit.com/raspberry-pi-i2c-clock-stretching-fixes/change-the-clock-speed

Three questions:

1. Why is this setting ignored, or, why can't it be changed in Python?
2. Can these lines be uncommented and changed to throw a Warning instead of ignoring the parameter silently?
3. If I can't changed the speed in Python, what should I do instead?

makermelissa:
Fixes #682. Also improves issue template messages.

### DIFF
--- a/src/adafruit_blinka/microcontroller/generic_linux/i2c.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/i2c.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: MIT
 """Generic Linux I2C class using PureIO's smbus class"""
+
+import warnings
 from Adafruit_PureIO import smbus
 
 
@@ -25,7 +27,9 @@ class I2C:
         _mode = self.MASTER
 
         if baudrate is not None:
-            print("I2C frequency is not settable in python, ignoring!")
+            warnings.warn(
+                "I2C frequency is not settable in python, ignoring!", RuntimeWarning
+            )
 
         try:
             self._i2c_bus = smbus.SMBus(bus_num)
--- a/src/adafruit_blinka/microcontroller/generic_linux/i2c.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/i2c.py
@@ -6,7 +6,11 @@ from Adafruit_PureIO import smbus
 
 
 class I2C:
-    """I2C class"""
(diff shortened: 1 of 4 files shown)


src/adafruit_blinka/microcontroller/generic_linux/i2c.py
```diff
@@ -20,8 +24,8 @@ def __init__(self, bus_num, mode=MASTER, baudrate=None):
             raise NotImplementedError("Only I2C Master supported!")
         _mode = self.MASTER
 
-        # if baudrate != None:
-        #    print("I2C frequency is not settable in python, ignoring!")
+        if baudrate is not None:
+            print("I2C frequency is not settable in python, ignoring!")
```
tannewt, line 28: [`warnings`](https://docs.python.org/3/library/warnings.html) is probably the best way to do it so folks can turn it off.

tannewt (changes requested):
One suggestion instead of print.

src/adafruit_blinka/microcontroller/generic_linux/i2c.py
makermelissa: Oh, I didn't know about this. Good idea.

tannewt (approved):
Thanks for the quick update!

PR #710, opened 2023-09-12
Update Usage Example in Readme

### LINKED ISSUE #600: Tests from Docs Fail
Maybe this is just a documentation update? I'm not seeing any testing module installed.

Example from [docs](https://docs.circuitpython.org/projects/blinka/en/latest/index.html#usage-example)
```
from testing import test_module_name
test_module_name("testing.universal.digitalio")
```
 Error
```
import testing
ModuleNotFoundError: No module named 'testing'
```

makermelissa:
Fixes #600. I replaced the usage example with something more practical.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,15 +77,27 @@ To install in a virtual environment in your current project:
 Usage Example
 =============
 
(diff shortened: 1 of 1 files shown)


jepler (approved):
thanks!

PR #713, opened 2023-09-14
Update busio to avoid always triggering frequency warning

makermelissa:
I realized the reason the warning was commented out before #708 is because busio uses a default frequency value of 100000 which will always trigger the warning. This will fix it so if the user isn't trying to set the frequency to something besides the default, it won't trigger the warning.

### DIFF
--- a/src/busio.py
+++ b/src/busio.py
@@ -138,6 +138,8 @@ class I2C(Lockable):
 
         if detector.board.any_embedded_linux:
             from adafruit_blinka.microcontroller.generic_linux.i2c import I2C as _I2C
+            if frequency == 100000:
+                frequency = None    # Set to None if default to avoid triggering warning
         elif detector.board.ftdi_ft2232h:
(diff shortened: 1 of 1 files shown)


caternuson (approved):

tannewt:
Is this the default speed for linux too? Can you read what speed linux is set to somehow?

caternuson:
> Is this the default speed for linux too?

yep

> Can you read what speed linux is set to somehow?

hmmm...maybe? here's a pi example:
```python
pi@raspberrypi:~ $ python3
Python 3.9.2 (default, Mar 12 2021, 04:06:34) 
[GCC 10.2.1 20210110] on linux
>>> import struct
>>> with open('/sys/bus/i2c/devices/i2c-1/of_node/clock-frequency','rb') as fp:
...     raw = fp.read()
... 
>>> print(raw)
b'\x00\x01\x86\xa0'
>>> struct.unpack(">I", raw)
(100000,)
>>>
```
not sure how universal that is across linux distros though

PR #715, opened 2023-09-15
Use latest release drafter to fix issues

makermelissa:
I'm running into this issue, which was fixed in a more recent release: https://github.com/release-drafter/release-drafter/issues/204

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -20,6 +20,6 @@ jobs:
         pull-requests: read
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


dhalbert (commented):
release-drafter/release-drafter@v5 should pick up the latest version, @v5.24.0, automatically. That version is tagged with @v5 also: https://github.com/release-drafter/release-drafter/tree/v5. So maybe you are using @v5.24.0 (>@v5.21.1), and it's not fixing the problem??

makermelissa:
Thanks, issue may have just cropped up again then.

PR #719, opened 2023-09-18
Update SPI to work with ReadableBuffers

### LINKED ISSUE #509: SPI writes do not accept NumPy buffers
Comparing to CircuitPython, the `write_xxx` functions should accept a `ReadableBuffer` as the buffer which on CircuitPython is:
* bytes _(standard Python)_
* bytearray _(standard Python)_
* memoryview _(standard Python)_
* array.array _(standard Python)_
* ulab.numpy.ndarray _(equates to numpy.ndarray)_
* rgbmatrix.RGBMatrix _(I don't believe there is an implementation for Blinka)_

```python
import board
import numpy

spi = board.SPI()

with spi:
    buffer = numpy.zeros(128, dtype="B")
    spi.write(buffer)
```

crashes with:
```python
  File "failure.py", line 8, in <module>
    spi.write(buffer)
  File ".env/lib/python3.7/site-packages/busio.py", line 349, in write
    return self._spi.write(buf, start, end)
  File ".env/lib/python3.7/site-packages/adafruit_blinka/microcontroller/generic_linux/spi.py", line 73, in write
    if not buf:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

makermelissa:
Fixes #509. Using a boolean truthy statement wasn't working with Numpy Buffers, but did work using tuples and such. Using len() works with all the types of buffers.

### DIFF
--- a/src/adafruit_blinka/microcontroller/generic_linux/spi.py
+++ b/src/adafruit_blinka/microcontroller/generic_linux/spi.py
@@ -73,7 +73,7 @@ class SPI:
 
     def write(self, buf, start=0, end=None):
         """Write data from the buffer to SPI"""
-        if not buf:
+        if buf is None or len(buf) < 1:
             return
         if end is None:
             end = len(buf)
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Thanks!

PR #722, opened 2023-10-07
Add Support for the Pcduino2 and Pcduino3

Ryzer58:
Tested pins D0- D6 using LEDs then repeating on pins D8 - D13.

I2C (i2c-dev) - tested with Adafruit PWM/Servo shield (PCA9685) using servokit library.

SPI (spidev) - Initial detection works. Tested with Ili9341 but was not able to get working, possibly an issue with the CS pin. The internal LED attached to clock pin flashes when trying to run examples under RGB Display. Hoping to find a fix in the near future.

UART - Tested with GPS module (SIM28 chipset) using CircuitPython GPS library on uart 2. Manages to pick up
satelites within a couple of seconds.

PWM - Added support for channel 0 to Allwinner A20 pin file. Tested with pulsing led based
on the example in the CircuitPython essentials guide.

Tested on Armbian 23.05 running kernel 6.1


makermelissa (approved):
Thank you. Looks good.

PR #728, opened 2023-10-23
Added Pi 5 Support + Fixed libgpiod pulls

makermelissa:
This adds support for the Raspberry Pi 5 as well as the Compute Module 4S (untested). RPi.GPIO is not currently working on the Pi 5 and since we already had a pin for libgpiod, I went with that. In order to maintain compatibility, I updated the code to fix the pullups and pulldowns, which should be helpful for other boards that use libgpiod.


tannewt (approved):
Looks good! Not tested.

makermelissa:
Fair enough. I'll do further testing and submit a followup PR if necessary.

PR #739, opened 2023-11-14
New Support for ODROID-M1S

how2flow:
i2c0, spi and uart are not available as default values because they are provided as aliases.

### DIFF
--- a/src/adafruit_blinka/board/hardkernel/odroidm1s.py
+++ b/src/adafruit_blinka/board/hardkernel/odroidm1s.py
@@ -1,0 +1,59 @@
+# SPDX-FileCopyrightText: 2023 Steve Jeong for Hardkernel
+#
+# SPDX-License-Identifier: MIT
+"""Pin definitions for the Odroid M1S."""
+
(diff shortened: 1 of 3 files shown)


how2flow:
+ https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/326

how2flow:
> Because you added a new microcontroller, they should be also added to: https://github.com/adafruit/Adafruit_Blinka/blob/main/src/microcontroller/__init__.py https://github.com/adafruit/Adafruit_Blinka/blob/main/src/microcontroller/pin.py

ODROID-M1S is rk3566 single board.
And rk3566 is already included in the code.


https://github.com/adafruit/Adafruit_Blinka/blob/4e29053a25b418de8f3f39fb6d4db9cfc6b23290/src/microcontroller/__init__.py#L108

https://github.com/adafruit/Adafruit_Blinka/blob/4e29053a25b418de8f3f39fb6d4db9cfc6b23290/src/microcontroller/pin.py#L114

how2flow:
Request review is my mistake..

PR #750, opened 2023-12-06
ODROIDs: Support aliased alt functions

how2flow:
Use rk3568 instead of rk3568b2:

- https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/331


makermelissa (approved):
Thanks. I just fixed a small formatting issue and now it's passing.

PR #751, opened 2023-12-08
To support pwm on ODROIDs

how2flow:
PlatformDetect:
- https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/333

### DIFF
--- a/src/adafruit_blinka/microcontroller/alias.py
+++ b/src/adafruit_blinka/microcontroller/alias.py
@@ -25,6 +25,7 @@ def get_dts_alias(device: str) -> Optional[str]:
                     return match.group(1).upper()
     return None
 
+
 def get_pwm_chipid(device: str):
     """Get the PWM Chip ID"""
     for chipid in range(32):
--- a/src/adafruit_blinka/microcontroller/alias.py
+++ b/src/adafruit_blinka/microcontroller/alias.py
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 """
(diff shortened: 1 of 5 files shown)


how2flow:
- https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/333
It seems that it's been two days since I left a comment, but there is no comment.
so I closed it

Anyway, This is a commit for support PWM on odroid boards.
Please review it.

Thanks.

makermelissa:
Hi, I had taken some time off and have been swamped ever since I got back, so reviews might be taking a bit longer than usual until I'm caught up. Please have some patience. Thanks.

makermelissa (approved):
I ran pre-commit as ithat was needed to get past the checks failing. Looks good to me. Thanks for doing this.

how2flow:
> Hi, I had taken some time off and have been swamped ever since I got back, so reviews might be taking a bit longer than usual until I'm caught up. Please have some patience. Thanks.

Thank you always for your hard work.

CircuitPython is a great platform.
But I only found out a month ago that I2C, SPI, PWM, etc. have not worked on Odroid for years in this good framework.
ODROID-M1 wasn't even working.

Well, not many people are using it, so I just found out,
but why don't many people use circuit python on Odroid?

Odroids are mainly used in certain products.
To be more loved for industrial or educational purposes, I thought it should be easier to control GPIO and various interfaces with C or Python.

We're almost there.
Odroid has 2 channels of ADC and none of them are working..

PR #756, opened 2023-12-15
generic_linux: pwmout: Remove duty-cycle reset in deinit flow

how2flow:
The 'deinit' function is called each time a property value is read and written.
However, the duty cycle continues to reset, so the sysfs PWM does not work properly.


makermelissa (approved):
Thanks. Looks good.

PR #761, opened 2023-12-20
ODROID-COMMON: Fix error of importing board

how2flow:
Character parsing index error.
Uart serial is aliased as SERIAL'X'
At the same time as fixing the error, I changed all the indexing to unify the format.


how2flow:
I don't think what I worked under 'site-packages' was reflected when it merged..
The index was unified at the same time as the modification.

I'm sorry, I'll check one more time before the PR.

makermelissa (approved):
Thanks for the update on this.

PR #763, opened 2023-12-26
microcontroller/rk: alias: Shorten the import statements to a single …

how2flow:
…line


how2flow:
If a new function is added and more than 3 modules are imported, it would be better to modify it to '*'.

makermelissa (approved):
Thanks.

PR #765, opened 2023-12-28
ODROID-COMMON: Support ADC pins

how2flow:
e.g.

```
import time
import board
from analogio import AnalogIn

analog_in = AnalogIn(board.A0) # support A0 and A1.

def get_voltage(pin):
    return (pin.value * 1.8) / 1024

while True:
    print((get_voltage(analog_in),))
    time.sleep(0.1)
```


makermelissa:
Please run pre-commit to reformat the files and then commit and push the reformatting.

makermelissa (approved):
Thank you. Looks good.

PR #769, opened 2024-01-03
ODROID-COMMON: Change pin mapping to support new dtbo and to unify with other odroid boards

how2flow:
This is to support default dtbo for circuitpython in kernel


how2flow:
+ (ODROID N2/C4) Add UART pins and Fix UART addresses

makermelissa (approved):
Thank you. Looks good.

PR #771, opened 2024-01-06
Fix ODROID_M1 board detection

mahenzon:
Yep, it worked because
```python
"abc" in "abc"  # is True
```
But I guess the idea was to check if it's one of the boards (and for now the tuple of boards has only one member)

Same as:
https://github.com/adafruit/Adafruit_Blinka/blob/fde264375432951da98131182e9b00a177b2f3c1/src/microcontroller/pin.py#L27

### DIFF
--- a/src/microcontroller/pin.py
+++ b/src/microcontroller/pin.py
@@ -4,7 +4,7 @@
 """Pins named after their chip name."""
 
 import sys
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks, makes sense.

PR #772, opened 2024-01-06
Add support for Repka Pi-3 H5 and Pi-4 H6 boards

mahenzon:
Repka-Pi 3 H5 differs from the standard H5 pinout. Add pinout to the lib

This PR depends on another PR in the PlatformDetect library https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/336
Pipelines will stay red until PlatformDetect is updated


makermelissa (approved):
Thank you. Looks good.

Issue #740, opened 2023-11-22
RPI 5 - python library piod: 'Chip' has no attribute 'OPEN_BY_NUMBER'


jasperjonker:
### Board Name

Raspberry Pi 5

### Steps

When following the following steps on the raspberry pi 5:

```py
pyenv virtualenv system test
pyenv shell test
pip install --upgrade Adafruit-Blinka
```
and then run blinka.py:
```py
import board
import digitalio
import busio

print("Hello blinka!")

# Try to great a Digital input
pin = digitalio.DigitalInOut(board.D4)
print("Digital IO ok!")

# Try to create an I2C device
i2c = busio.I2C(board.SCL, board.SDA)
print("I2C ok!")

# Try to create an SPI device
spi = busio.SPI(board.SCLK, board.MOSI, board.MISO)
print("SPI ok!")

print("done!")
```

I get the following error:
```sh
ImportError: libgpiod Python bindings not found, please install and try again! See https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/master/libgpiod.sh
```

Installing gpiod: `pip install --upgrade gpiod`
```sh
❯ python blinka.py
Traceback (most recent call last):
  File "/home/utb/Downloads/blinka.py", line 1, in <module>
    import board
  File "/home/utb/.pyenv/versions/test2/lib/python3.11/site-packages/board.py", line 47, in <module>
    from adafruit_blinka.board.raspberrypi.raspi_5b import *
  File "/home/utb/.pyenv/versions/test2/lib/python3.11/site-packages/adafruit_blinka/board/raspberrypi/raspi_5b.py", line 6, in <module>
    from adafruit_blinka.microcontroller.bcm2712 import pin
  File "/home/utb/.pyenv/versions/test2/lib/python3.11/site-packages/adafruit_blinka/microcontroller/bcm2712/pin.py", line 8, in <module>
    D0 = Pin((4, 0))
         ^^^^^^^^^^^
  File "/home/utb/.pyenv/versions/test2/lib/python3.11/site-packages/adafruit_blinka/microcontroller/generic_linux/libgpiod_pin.py", line 36, in __init__
    self._chip = gpiod.Chip(str(pin_id[0]), gpiod.Chip.OPEN_BY_NUMBER)
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'Chip' has no attribute 'OPEN_BY_NUMBER'
```

### Description

[...]

Issue #685, opened 2023-06-12
libgpiod 1.4.2 or older results in error: '<' not supported between instances of 'builtin_function_or_method' and 'int'


how2flow:
If the Using libgpiod version is low, the following error might occur. (I am using libgpiod v1.4.2.)
```
...

'<' not supported between instances of 'builtin_function_or_method' and 'int'
```
Logs that appear after https://github.com/adafruit/Adafruit_Blinka/pull/678 is reflected, i.e. when you use pip package version 8.19.0 or later.
Run command `sed -i s/num_lines/num_lines\(\)/g {your_prefix}/microcontroller/amlogic/meson_g12_common/pin.py` or Change the version to 8.18.1 or lower.
Fixing a simple error line may have some side effects. I recommend that you lower the version.

I leave a message for those who meet the error log.

makermelissa:
It appears an attempt was made to update this file to work with both older and newer versions of libgpiod as evidenced by snippets of code like:
```
if hasattr(gpiod, "Chip"):
    chip0 = gpiod.Chip("0")
    chip1 = gpiod.Chip("1")
else:
    chip0 = gpiod.chip("0")
    chip1 = gpiod.chip("1")
```

However, it appears that some of the logic may have been missed.

PR #784, opened 2024-02-06
libgpio 2.x support for odroid c4

janvolck:
After an upgrade of my python environment I got issues with libgpiod

A quick search learned me that libgpiod has been updated again with breaking changes
The same issue was already faced with RPI5 in issue https://github.com/adafruit/Adafruit_Blinka/issues/740

I much like the clean and neat approach taken with libgpiod_pin and made an attempt to mimic the same in libgpiod_chip for supporting multiple libgpiod versions on Odroid C4 (meson_g12_common).

Do note that in the past I created some regression causing issue https://github.com/adafruit/Adafruit_Blinka/issues/685
All this logic is now in libgpiod_chip.py, so I hope it didn't break anything. As @how2flow is also very active lately I hope he might also have a peak at it.

This approach could also be used Adafruit_Blinka/src/adafruit_blinka/microcontroller/amlogic/a311d/pin.py but as I can't test this I didn't touch it.


makermelissa (approved):
Thanks. Looks good to me.

how2flow:
@janvolck, Thank you for your hard work

C4 and N2 share the same module (meson_g12_common), so I should test it together! Thank you.
Also, because it is a module under generic_linux, it will be solved in other models.

Issue #776, opened 2024-01-11
Raspberry Pi 5 PWMOut needs updating to libgpiod


makermelissa:
### Board Name

Raspberry Pi 5

### Steps

1. Install Blinka
2. Install ST7789 lib
3. Run bonnet example

### Description

The following error occurs.
```
Traceback (most recent call last):
  File "/home/pi/st7789_240x240_bonnet_simpletest.py", line 33, in <module>
    display = ST7789(
              ^^^^^^^
  File "/home/pi/env/lib/python3.11/site-packages/adafruit_st7789.py", line 72, in __init__
    super().__init__(bus, _INIT_SEQUENCE, **kwargs)
  File "/home/pi/env/lib/python3.11/site-packages/displayio/_display.py", line 213, in __init__
    self._backlight = PWMOut(backlight_pin, frequency=100, duty_cycle=0)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pi/env/lib/python3.11/site-packages/adafruit_blinka/microcontroller/bcm283x/pwmio/PWMOut.py", line 27, in __init__
    self._open(pin, duty_cycle, frequency, variable_frequency)
  File "/home/pi/env/lib/python3.11/site-packages/adafruit_blinka/microcontroller/bcm283x/pwmio/PWMOut.py", line 40, in _open
    GPIO.setup(pin.id, GPIO.OUT)
RuntimeError: Cannot determine SOC peripheral base address
```

The issue is that version of PWMOut uses RPi.GPIO, which is not available on the Pi 5. The PWMOut for the bcm2712 needs to be rewritten to use libgpiod.

makermelissa:
It looks like the library lgpio (https://github.com/waveform80/rpi-lgpio) may support pwm and run on the pi 5. I'm going to see if it can peacefully coexist with gpiod.

PR #785, opened 2024-02-08
Show NotImplementedError for pwmio on the Pi 5 for now

makermelissa:
PWMIO is not working on the Pi 5 at the moment (see #776). This is more of a stopgap measure to at least show a more appropriate error until it has been fixed.


dhalbert (approved):
:+1:

PR #790, opened 2024-02-28
ODROID-M1/S: Fix that pwm1, pwm2 pins do not match with kernel dtbos

how2flow:


how2flow:
anyone..?

makermelissa (approved):
Thanks. Sorry for the delay. It looks like a couple Lubancat boards use the RK3566, but don't have PWM implemented, so this change should be fine.

PR #801, opened 2024-03-22
add quartz64a

anonymouscowhead:
Add support for the Quartz64 model A board from Pine64


anonymouscowhead:
weird pylint issue.. I think it is because there is a pine64.py (for the old board of that name) alongside my new pine64 directory (for the vendor).
Guess tomorrow I'll try moving the existing board file under my dir...

makermelissa (approved):
Looks great. Thank you.

PR #807, opened 2024-04-02
Add support for "fake" MCP2221

brentru:
This pull request adds Blinka support for a "fake" MCP2221 device. This means that no device is attached and the environment will mock the CircuitPython hardware API layer.

To test, set the following environment variables in either your shell or a CPython script:
```
os.environ["BLINKA_FORCEBOARD"] = "MICROCHIP_MCP2221"
os.environ["BLINKA_ FORCECHIP"] = "MCP2221"
```

Implemented and tested on CPython: 
* Digital Input
  * Returns a True/False value, toggles between them on each call
* Analog Input
  * Returns an arbitrary 16-bit value
* I2C
  * Scan (returns reasonable values, otherwise can feed the scan() function it a list of arguments as a test stimuli) 
  * Support for ["Fake BME280" driver](https://github.com/adafruit/Adafruit_CircuitPython_Fake_BME280/)
    * This driver mocks the functionality of the [Adafruit_CircuitPython_BME280](https://github.com/adafruit/Adafruit_CircuitPython_BME280), allowing you to test your code without attaching physical hardware. Instead of using random data, weather data is instead pulled from [the OpenWeatherMaps API](https://openweathermap.org/) and returned as if it were coming from the BME280 sensor properties.


src/adafruit_blinka/microcontroller/fake_mcp2221/i2c.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class I2C:
+    """Custom I2C Class for MCP2221"""
+
+    def __init__(self, *, frequency=100000):
+        self._mcp2221 = mcp2221
+        self._freq = frequency
+
+    def scan(address_list=None):
```
makermelissa, line 16: You might want to add the @staticmethod decorator here

brentru: @makermelissa Added!

src/board.py
```diff
@@ -224,7 +224,10 @@
     from adafruit_blinka.board.binho_nova import *
 
 elif board_id == ap_board.MICROCHIP_MCP2221:
-    from adafruit_blinka.board.microchip_mcp2221 import *
+    if "BLINKA_FORCEBOARD" not in os.environ:
```
makermelissa, line 227: Can we change this to something like
```
    if "BLINKA_FORCECHIP" in os.environ and os.environ["BLINKA_FORCEBOARD"] == "MICROCHIP_MCP2221":
        from adafruit_blinka.board.fake_microchip_mcp2221 import *
    else:
        from adafruit_blinka.board.microchip_mcp2221 import *
```
Because this lies on an outside package, the possibility of os.environ["BLINKA_FORCEBOARD"] being set to something else but board_id being equal to ap_board.MICROCHIP_MCP2221 is extremely improbable, but not impossible.

src/busio.py
```diff
@@ -50,8 +51,10 @@ def init(self, scl, sda, frequency):
             self._i2c = _I2C(frequency=frequency)
             return
         if detector.board.microchip_mcp2221:
-            from adafruit_blinka.microcontroller.mcp2221.i2c import I2C as _I2C
-
+            if "BLINKA_FORCEBOARD" not in os.environ:
```
makermelissa, line 54: Same here

src/digitalio.py
```diff
@@ -127,7 +127,10 @@
 elif detector.board.greatfet_one:
     from adafruit_blinka.microcontroller.nxp_lpc4330.pin import Pin
 elif detector.board.microchip_mcp2221:
-    from adafruit_blinka.microcontroller.mcp2221.pin import Pin
+    if "BLINKA_FORCEBOARD" not in os.environ:
```
makermelissa, line 130: Same here

src/microcontroller/__init__.py
```diff
@@ -123,7 +123,10 @@ def delay_us(delay):
 elif chip_id == ap_chip.LPC4330:
     from adafruit_blinka.microcontroller.nxp_lpc4330 import *
 elif chip_id == ap_chip.MCP2221:
-    from adafruit_blinka.microcontroller.mcp2221 import *
+    if "BLINKA_FORCECHIP" not in os.environ:
```
makermelissa, line 126: Same here

src/microcontroller/pin.py
```diff
@@ -97,7 +97,11 @@
 elif chip_id == ap_chip.LPC4330:
     from adafruit_blinka.microcontroller.nxp_lpc4330.pin import *
 elif chip_id == ap_chip.MCP2221:
-    from adafruit_blinka.microcontroller.mcp2221.pin import *
+    if "BLINKA_FORCECHIP" not in os.environ:
```
makermelissa, line 100: Same here

src/analogio.py
```diff
@@ -9,16 +9,20 @@
...
 import sys
 
 from adafruit_blinka.agnostic import detector
 
 # pylint: disable=ungrouped-imports,wrong-import-position,unused-import
 
 if detector.board.microchip_mcp2221:
-    from adafruit_blinka.microcontroller.mcp2221.analogio import AnalogIn
-    from adafruit_blinka.microcontroller.mcp2221.analogio import AnalogOut
+    if "BLINKA_FORCECHIP" not in os.environ:
```
makermelissa, line 20: Same as in board.py

makermelissa (changes requested):
There are a bunch of code blocks where it leaves the slimmest possibility of it not functioning as intended. Other than that, it looks good.

makermelissa (approved):
Looks great. Thank you.

PR #809, opened 2024-04-05
Fix PWMIO import bug

### LINKED ISSUE #808: Bug in pwmio.py introduced with commit e13c0662ce0da9e936c652953aa1d3f9d6b872d5
### Board Name

Raspberry Pi 4

### Steps

1. Install current Adafruit_Blinka via pip3
2. Start python shell and try to import pwmio
3. Get error:
```
>>> import pwmio
  File "/home/pi/lf7/.venv/lib/python3.11/site-packages/pwmio.py", line 62, in <module>
    raise NotImplementedError("pwmio not supported for this board.")
NotImplementedError: pwmio not supported for this board.
```

### Description

_No response_

### Additional information

Line 45 in src/pwmio.py should probably be `elif` instead of `if`. In its current form importing pwmio raises an error even if any detection above that line returns True.
https://github.com/adafruit/Adafruit_Blinka/blob/e13c0662ce0da9e936c652953aa1d3f9d6b872d5/src/pwmio.py#L45-L46

makermelissa:
Fixes #808.

### DIFF
--- a/src/pwmio.py
+++ b/src/pwmio.py
@@ -42,7 +42,7 @@ elif detector.board.any_lubancat:
     from adafruit_blinka.microcontroller.generic_linux.sysfs_pwmout import PWMOut
 elif detector.board.pico_u2if:
     from adafruit_blinka.microcontroller.rp2040_u2if.pwmio import PWMOut
-if detector.board.any_jetson_board:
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks for the quick fix!

PR #817, opened 2024-04-22
Add Raxda ZERO 3 support

burberius:
I tested digital output on all pins and also I2C, for SPI I don't have a device currently.


vinicentus:
The checks should probably pass now that https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/348 is merged?

vinicentus:
Can confirm that I2C works on both I2C3 (GPIO1_A0 and GPIO1_A1) and I2C4 (GPIO4_B2, GPIO4_B3). I'm able to communicate with a device using both interfaces.

makermelissa (approved):
Thank you. Looks good.

PR #749, opened 2023-12-06
Fix pin number assignment for allwinner h616

lz00qs:
Link to https://github.com/adafruit/Adafruit_Blinka/issues/689

Modify the format of the pin parameter to fix errors bellow:

```
Traceback (most recent call last):
  File "/root/oled.py", line 13, in <module>
    btn = DigitalInOut(board.PC11)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/digitalio.py", line 183, in __init__
    self.direction = Direction.INPUT
    ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/digitalio.py", line 213, in direction
    self._pin.init(mode=[Pin.IN](http://pin.in/))
  File "/usr/lib/python3.11/site-packages/adafruit_blinka/microcontroller/generic_linux/libgpiod_pin.py", line 56, in init
    self._line = self._chip.get_line(int(self._num))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 22] Invalid argument
```

Tested on orange pi zero2, archlinux, 6.1.31 orange pi official kernel.

PR #830, opened 2024-05-22
add support for walnutpi-1b board

sc-bin:
hello,we want to add support for the walnutpi-1b board -> [walnutpi.com](https://walnutpi.com/docs/intro/)

about the commit [modify pin.py for h616 : auto choose gpiochip](https://github.com/adafruit/Adafruit_Blinka/pull/830/commits/2a3697f4f1bc3e9b4cb55b49236d445c82593fcb)
I have used kernel version 5.16 and 6.1 on our board , run the gpiodetect ,That's always the result
```
pi@WalnutPi:~$ gpiodetect
gpiochip0 [300b000.pinctrl] (288 lines)
gpiochip1 [7022000.pinctrl] (32 lines)
```
But it seems to be on some H616 board, the  [300b000.pinctrl] is belong th gpiochip1。-> [the pull #749 ](https://github.com/adafruit/Adafruit_Blinka/pull/749#issuecomment-1970728186)

so i modfly the pin.py , it will read the system file `/sys/class/gpio/gpiochip0/label` ,if the label for gpiochip0 is [300b000.pinctrl] ,then all the pins will Be registered as 0 ,If not, register as 1.


makermelissa (approved):
Thank you.

PR #828, opened 2024-05-21
Add "generic os-agnostic" board and remove "fake_mcp2221" board

brentru:
This PR adds support for a generic, OS-agnostic board for broader use cases (such as with pytest). It also removes the "fake_mcp2221" board and its related files, since the new "generic agnostic board" is intended to supersede this board.

@lorennorman  You may want to take a look/review pass over this as well since we will switch to using it in WipperSnapper Python.

* Added new board, `generic_agnostic_board`, with support for `digitalio`, `analogio`, and `busio` modules.
  * This board includes "expected pin behaviors" such as returning expected pin values for `digitalio` input and output, `analogio` input and output, and special analog pin functions like waveforms.
* Added new tests/ directory with pytest files for this board
* Deleted files related to the `fake_mcp2221` board

Test Results:
```
pytest tests/ -v                             906ms  Tue May 21 14:40:07 2024
========================================================= test session starts ==========================================================
platform darwin -- Python 3.12.2, pytest-8.2.1, pluggy-1.5.0 -- /Users/brentrubell/Desktop/github_brentru/Adafruit_Blinka/.env/bin/python3.12
cachedir: .pytest_cache
rootdir: /Users/brentrubell/Desktop/github_brentru/Adafruit_Blinka
collected 12 items

tests/test_generic_agnostic_board_analogio.py::test_Ax_OUTPUT PASSED                                                             [  8%]
tests/test_generic_agnostic_board_analogio.py::test_Ax_INPUT_RAND_INT PASSED                                                     [ 16%]
tests/test_generic_agnostic_board_analogio.py::test_Ax_INPUT_FIXED_INT_PI PASSED                                                 [ 25%]
tests/test_generic_agnostic_board_analogio.py::test_Ax_INPUT_WAVE_SINE PASSED                                                    [ 33%]
tests/test_generic_agnostic_board_analogio.py::test_Ax_INPUT_WAVE_SAW PASSED                                                     [ 41%]
tests/test_generic_agnostic_board_digitalio.py::test_Dx_OUTPUT_TRUE PASSED                                                       [ 50%]
tests/test_generic_agnostic_board_digitalio.py::test_Dx_INPUT_TRUE PASSED                                                        [ 58%]
tests/test_generic_agnostic_board_digitalio.py::test_Dx_INPUT_TRUE_PULL_DOWN PASSED                                              [ 66%]
[...]

PR #831, opened 2024-05-24
Add SPI, NeoPixel to "generic os-agnostic" board

brentru:
This pull request adds the following to get the CircuitPython NeoPixel and DotStar libraries working with the "generic os-agnostic" board:
* Mock SPI class
* Mock `neopixel_write.py`
* Adds descriptive pins for use with pixels to generic os-agnostic board pinout, `board.NEOPIXEL`, `board.DOTSTAR_DATA`, `board.DOTSTAR_CLK`
* Adds `spiPorts` to generic os-agnostic board pinout

Related: https://github.com/adafruit/Adafruit_Blinka/pull/828


makermelissa (changes requested):
I'd like to keep environ checks to just be in platformdetect, which should should be returning a generic agnostic chip and board. Then the result of the platformdetect should be used here.

PR #835, opened 2024-06-01
Added support for Radxa Rock 5C

hajimef:
I added support for Radxa Rock 5C.
I aso added support for Radxa Rock 5C to Adafruit_Python_Platformdetect .


makermelissa (approved):
Looks good. Thank you.

PR #837, opened 2024-06-03
Fix SPI on generic board

brentru:
Fixes SPI configuration for the generic OS board by adding extra detection to to the `busio` `configure()` method


makermelissa:
Looks like a formatting issue. I'll add this in a bug fix release after it passes.

brentru:
@makermelissa  I ran black across the file, should be good to go after the test passes

makermelissa (approved):
Thank you. Looks good.

PR #846, opened 2024-06-13
added orangepipc2 board

nopnop2002:
added orangepipc2 board to blinka


makermelissa (approved):
Thank you. Looks great.

PR #850, opened 2024-06-14
Add PWMOut Support for Generic Agnostic Board

brentru:
`PWMOut` implementation for the Generic Agnostic Board. For this board, the `PWMOut` class is a mock PWM output interface for use with software (CI) testing.


makermelissa (approved):
Looks good. Thanks.

PR #853, opened 2024-06-18
Add support for lubancat-5

mmontol:
In the 40PIN of Lubancat-5, some GPIOs belong to the extended IO and not to the rk3588 chip, so the pins only have IO functions and cannot be reused for other functions.  I choose to add `from adafruit_blinka.microcontroller.generic_linux.libgpiod_pin import Pin` to the Lubancat5. py to control these IOs.


makermelissa (approved):
Looks good. Thanks.

PR #873, opened 2024-07-23
Remove unused numpy as requirement

### LINKED ISSUE #872: Remove dependency to numpy
Please remove the dependency to `numpy` package.  It is an insanely bloated package (20MB compared to 200-300 kB for other packages) and I'm having tremendous problems installing it on a Raspi Zero W (installing it with poetry fails due to many compilation problems and I finally hit a brick wall I'm unable to solve, and installing using pip also requires e.g. g++ to be installed and takes tens of minutes).

The [only reference I can find to numpy](https://github.com/search?q=repo%3Aadafruit%2FAdafruit_Blinka+numpy&type=code) is the dependency, I can't see it used in any code.  numpy is for scientific calculation, I cannot see why a package like Adafruit blinka would need it.

Thanks.

makermelissa:
Fixes #872. Numpy seems to no longer be needed as of #552.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,6 @@ setup(
         "Adafruit-PureIO>=1.1.7",
         "binho-host-adapter>=0.1.6",
         "pyftdi>=0.40.0",
-        "numpy>=1.21.5",
         "adafruit-circuitpython-typing",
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Great!

Issue #881, opened 2024-08-11
[Suggestion] Find incompatibilities w.r.t. CircuitPython implementation


elpekenin:
Yesterday, a user reported on discord an issue in which their library's CI was failing.

Upon a quick investigation, i found out that it had to do with slight differences between the CircuitPython code they were running on their board and the Blinka code that was later running/failing on CI.

Since we have both
* The implementation (Blinka)
* The typing stubs it should[1] match with (circuitpython-stubs)

...some automated testing could be run comparing them, to have lists of "this is not implemented yet" or "this function doesnt work the same way". 

I ran a very quick Google search and seen that MyPy's `stubtest` command could be used (there might be other/better alternatives out there).

Since im not familiar with the codebase nor said tool, I havent really tested anything but thought it was a good thing to share the idea in case you guys find it useful.

---

[1] Im not sure whether 100% compatibility is desired or even posible on all cases... (not talking about coverage, but about function signatures and whatnot)

makermelissa:
Yeah, they should be the same where the functionality exists So if a module is present in Blinka, code should run with very little modification (Pin names are ok to differ). So for function signatures, you are correct. We try and match, but occasionally CircuitPython makes changes that I'm not aware of. Do you know which function it was?

Also, I like you idea of a tool that we can use to compare that can be run during CI, so that we can be aware of those changes even without notification.

elpekenin:
The specific problem was `keypad`'s `EventQueue` being named `_EventQueue` instead

Will try and get some testing working locally

makermelissa:
Ah, gotcha. Looks like it was just fixed by #883. Will leave this open for a more general issue.

elpekenin:
My idea would be something like this: https://github.com/elpekenin/Adafruit_Blinka/tree/stubtest

elpekenin:
Some diagnostics after a couple (already-pushed) changes:
```py
Running for: BLINKA_OS_AGNOSTIC
===============================
analogio failed to import. NotImplementedError: analogio not supported for this board.
keypad.Event.__init__ is inconsistent, runtime does not have argument "timestamp"
keypad.Event.key_number is read-only at runtime but not in the stub
keypad.Event.pressed is read-only at runtime but not in the stub
keypad.Event.released is read-only at runtime but not in the stub
keypad.EventQueue.__init__ is inconsistent, stub does not have argument "max_events"
keypad.EventQueue.overflowed is read-only at runtime but not in the stub
[...]

PR #884, opened 2024-08-13
[Tooling] Initial implementation of stub checker

elpekenin:
As per title, just wrote a minimal (and very subject to improvements) script that will compare the actual implementation (`.py`) against the circuitpython-generated stubs (`.pyi`)

Note:
- So far, only runs against the `BLINKA_OS_AGNOSTIC` target
  - I havent tested other targets because i expect them to fail due to lack of some connected hardware, or my computer not being what blinka is being told using the `BLINKA_FORCECHIP` environment variable
- I dont know whether circuitpython's stubs (which im guessing are generated from the doc-comments on source code) are tested/validated in any way against the actual code that ends up exposed to Python's VM

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,3 @@
+# compare blinka's implementation vs the stubs coming from CircuitPython's implementation
+circuitpython-stubs
+mypy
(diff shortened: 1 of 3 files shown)


elpekenin:
Some more info in #881

PR #888, opened 2024-08-30
I2C writeto_then_readfrom in a single transaction on u2if

ryanmesh-sd:
Implemented WRITE_THEN_READ on I2C interface in U2IF (associated PR: https://github.com/adafruit/u2if/pull/12)

This allows reading from specific registers in I2C devices without sending two separate commands from the host, which significantly increases latency between the pointer write and read and causes timeouts in certain I2C devices (like the ones I've been using).


ryanmesh-sd:
Looks to now pass tests on my branch CI. Would it be better to do this with a fallback to the old (two separate HID reports) method to accommodate U2IF devices using older firmware?

FoamyGuy:
> Looks to now pass tests on my branch CI. Would it be better to do this with a fallback to the old (two separate HID reports) method to accommodate U2IF devices using older firmware?

Yes, I think it would be best for this to have fallback for devices with the older firmware. 

Thanks for working on this!

ryanmesh-sd:
Hi @FoamyGuy, I've added the fallback. Let me know what you think!

PR #894, opened 2024-09-17
add support for orangepi 3b

landall:


landall:
can anyone tell me why the test is not passed?
I do not understand the error message.

tannewt:
The easiest thing to do is run pre-commit locally. It'll automatically fix the formatting. Instructions are here: https://learn.adafruit.com/improve-your-code-with-pylint/check-your-code-with-pre-commit

landall:
> The easiest thing to do is run pre-commit locally. It'll automatically fix the formatting. Instructions are here: https://learn.adafruit.com/improve-your-code-with-pylint/check-your-code-with-pre-commit

Fail to install pre-commit in my Win10.

pre-commit run --all-files
[INFO] Installing environment for https://github.com/fsfe/reuse-tool.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('C:\\Users\\CST\\AppData\\Local\\Microsoft\\WindowsApps\\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\\python.exe', '-mvirtualenv', 'C:\\Users\\CST\\.cache\\pre-commit\\repo5_yrkgf1\\py_env-python3')
return code: 1
stdout:
    RuntimeError: failed to build image setuptools because:
      File "C:\Users\CST\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\virtualenv\seed\embed\via_app_data\via_app_data.py", line 57, in _install
        installer.install(creator.interpreter.version_info)
      File "C:\Users\CST\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\virtualenv\seed\embed\via_app_data\pip_install\base.py", line 37, in install
        self._sync(filename, into)
      File "C:\Users\CST\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\virtualenv\seed\embed\via_app_data\pip_install\copy.py", line 13, in _sync
        copy(src, dst)
      File "C:\Users\CST\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\virtualenv\util\path\_sync.py", line 41, in copy
        method(str(src), str(dest))
      File "C:\Users\CST\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\virtualenv\util\path\_sync.py", line 52, in copytree
        shutil.copy(src_f, dest_f)
      File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\shutil.py", line 417, in copy
        copyfile(src, dst, follow_symlinks=follow_symlinks)
      File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\shutil.py", line 254, in copyfile
        with open(src, 'rb') as fsrc:
    FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\CST\\AppData\\Local\\Packages\\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\\LocalCache\\Local\\pypa\\virtualenv\\wheel\\3.10\\image\\1\\CopyPipInstall\\setuptools-75.1.0-py3-none-any\\pkg_resources\\tests\\data\\my-test-package_zipped-egg\\my_test_package-1.0-py3.7.egg'
stderr: (none)

landall:
Can anyone replace me to commit this code?
I really cannot handle this checks without any useful information.

FoamyGuy:
I ran pre-commit and committed the resulting changes. 

I'm not sure what the root cause of the error you posted is. If you have the option, and you still want to get it working locally it may be worth trying a newer python version like 3.11 or 3.12.

landall:
> I ran pre-commit and committed the resulting changes.
> 
> I'm not sure what the root cause of the error you posted is. If you have the option, and you still want to get it working locally it may be worth trying a newer python version like 3.11 or 3.12.

thanks very much.
I will try to build a working envirnment later in a clean machine.
Is there any way to print out the wrong lines in github checks page? I can see the hints directly when i commit to some other projects.

landall:
The second question: is there any guide when I commit to board / microcontroller definition?
I find the defined vars is different among these files.
I do not know which should be defined in a board/microcontroller files of adafruit_blinka.

landall:
Remark:
need reviewing the modification of i2cPorts in rk3566.py.
/board/radxa/radxacm3.py seems to use I2C2_*_M0 but /board/orangepi/orangepi3b use I2C2_*_M1.

src/adafruit_blinka/microcontroller/rockchip/rk3566/pin.py
```diff
@@ -169,7 +178,8 @@
 # ordered as i2cId, SCL, SDA
 i2cPorts = [
     (1, I2C1_SCL, I2C1_SDA),
-    (2, I2C2_SCL_M0, I2C2_SDA_M0),
+    # (2, I2C2_SCL_M0, I2C2_SDA_M0),
+    (2, I2C2_SCL_M1, I2C2_SDA_M1),
```
FoamyGuy, line 182: I think this likely to cause issues for other devices that were using the I2C bus. I think we would want to selectively make this change only for devices that need it. 

You could add another branch to the if block down here: https://github.com/adafruit/Adafruit_Blinka/blob/1c4358371c4945d08fb208fa83e425f5b1af9d45/src/adafruit_blinka/microcontroller/rockchip/rk3566/pin.py#L199 to detect the relavent board(s) and then when found manipulate the `i2cPorts` list as needed for that device.

FoamyGuy (changes requested):
I think we would want to find a way to make the changes to the rk3566/pin.py file compatible with all devices instead of commenting out the old one for the new one.

FoamyGuy:
I noticed in the pins file for the rk3588 that there are sometimes multiple ports listed under the same numbers: https://github.com/adafruit/Adafruit_Blinka/blob/9441d12d5313c1e62a103bcfbd2fb60cf354e4d7/src/adafruit_blinka/microcontroller/rockchip/rk3588/pin.py#L238-L241 Maybe the rest of the code already handles this situation without the need for a special case.

@landall can you try just including the new port along with the existing ones without commenting anything out like this:

```python
    (1, I2C1_SCL, I2C1_SDA),
    (2, I2C2_SCL_M0, I2C2_SDA_M0),
    (2, I2C2_SCL_M1, I2C2_SDA_M1),
```

See if that works on your orangepi 3b device and if so add a commit to uncomment the existing definition and add the new one without removing anything.

landall:
> I noticed in the pins file for the rk3588 that there are sometimes multiple ports listed under the same numbers:
> 
> https://github.com/adafruit/Adafruit_Blinka/blob/9441d12d5313c1e62a103bcfbd2fb60cf354e4d7/src/adafruit_blinka/microcontroller/rockchip/rk3588/pin.py#L238-L241
> 
> Maybe the rest of the code already handles this situation without the need for a special case.
> @landall can you try just including the new port along with the existing ones without commenting anything out like this:
> 
> ```python
>     (1, I2C1_SCL, I2C1_SDA),
>     (2, I2C2_SCL_M0, I2C2_SDA_M0),
>     (2, I2C2_SCL_M1, I2C2_SDA_M1),
> ```
> 
> See if that works on your orangepi 3b device and if so add a commit to uncomment the existing definition and add the new one without removing anything.

It works!
I commit it to this pull request now.

src/adafruit_blinka/board/orangepi/orangepi3b.py
```diff
@@ -0,0 +1,84 @@
+# SPDX-FileCopyrightText: 2021 Melissa LeBlanc-Williams for Adafruit Industries
```
FoamyGuy, line 1: Can you update the copyright to your name or username, and the year to 2024. You can leave of the "for Adafruit Industries" as well. 

This is looking good to me no beyond that.

landall: done.

FoamyGuy (approved):
Thank you!

Issue #910, opened 2024-11-13
Configuration Management problems with the use of rpi-gpio "hack"; just use lgpio native instead?


frank-pet:
### Board Name

Raspberry PI 5

### Steps

A pip update of Adafruit_blinka from a version earlier than 0.8.28(?) to any later version does not work.  This is due to the fact that the rpi-gpio replacement for RPi.GPIO is only installed from the ./adafruit/Raspberry-Pi-Installer-Scripts/master/raspi-blinka.py installer.  It took quite a while to track this down.

### Description

I'd like to propose an alternate solution.  Since the /dev/gpiochip* files and lgpio python package are now part of the Debian kernel 1:6.1.63-rpi7-rpi-v8, simply implement the ./microcontroller/bcm283x/pin.py and ./microcontroller/bcm283x/pwmio/PWMOut.py using lgpio directly, and remove the RPi.GPIO code (3 lines) from the ./microcontroller/bcm2711/pin.py package.   An updated version is available in pull request 911 and has been tested successfully on a PI5 on this kernel and a PI3B as well, using the pin package directly and also using digitalio.DigitalInOut with no changes to the digitalio package.   Also successfully tested the blinka pwmio.PWMOut package. on both platforms.

### Additional information

Code changes in pull request 911

frank-pet:
Hmmm, didn't expect to see the code formatted that way; I can just send you the file if you wish.

makermelissa:
Yeah, this is something I've considered. It would certainly solve one of the installation issues with the RPi.GPIO being installed and causing a conflict. The bcm2711 folder is also used by the Pi 4, which I'm not sure if lgpio would run on, so it may just need a bcm2712 version for the Pi 5 if it doesn't work on the Pi 4.

frank-pet:
I haven’t tested these change on a PI 4, but I don’t see why it wouldn’t work.  The kernel  on my bullseye install shows the /dev/gpio chip devices, so it should run just like it does on the PI3. The conflict you mentioned is a real mess to deal with.   Just had to deal with it again updating a Pi zero w.

makermelissa:
Fixed by #911

PR #911, opened 2024-11-16
Lgpio native

frank-pet:
Code changes to remove use of RPi.GPIO and replace with using the lgpio package directly.   changes to:

1. microcontroller/bcm2711/pin.py
2. microcontroller/bcm283x/pin.py
3. microcontroller/bcm283x/pwmio/PWMOut.py

Changes have been tested and work on both a PI 5 and a Pi3b.

This is the fix for Issue #910


makermelissa:
Overall, the changes in the PR look great and it appears you have done most of the hard stuff.

However, the bcm283x files are used by every single raspberry pi from Pi 1 through Pi 3B+ and I'd prefer to leave these alone as they seem to be fine. The bcm2811 files are used by Pi 4 and up and are probably safe to mess with. Because lgpio appears to be targeted more at linux and less at Raspberry Pis specifically, I would prefer to add new lgpio-specific pinfile to `src/adafruit-blinka/microcontroller/generic_linux` (something like lgpio_pin.py) so that they could be used by other boards in the future and maybe a copy of the PWMOut (with your changes) to the bcm2811 folder. Then only the bcm2811 files could just make use of those.

makermelissa:
Also, if you wanted to test with the Raspberry Pi Zero (original), that would probably be pretty equivalent to the Original Raspberry Pi.

frank-pet:
Just finished testing on a Pi zero w v1.1.   changes work just fine.  Took some effort since all my other projects are 64-bit machines; had to build a 32 bit image for this test...

frank-pet:
That was a while ago, but probably because I didn’t want to stray too far from the other code in the package I was looking at…. Yours does sound like a better approach.Sent from my iPadOn Feb 5, 2025, at 5:07 PM, Melissa LeBlanc-Williams ***@***.***> wrote:﻿
Thank you. The main odd thing in your code that I see is:
In PWMOut, rather than importing board, why not have:
from adafruit_blinka.microcontroller.bcm283x.pin import Pin, CHIP
And then you can remove the unused import from bcm2711.
I want to do some testing myself (such as speed tests) and if it is similar (or faster). I may do some restructuring as well so that other boards could use the lgpio pin def as well. It will at least allow us to have some consistency this way.

makermelissa:
I made some structural changes that I'm testing out today and if all looks good, I'll merge my changes to your PR and then merge this one.

makermelissa:
I restructured so that the Pi 3 and 4 are using RPi.GPIO for now and the Pi 5 is using lgpio. I tested on the Pi 3, Pi 4, and Pi 5 with basic digitalio, running Blinka Test, a PWM test, and importing microcontroller. For the Pi 5, I made sure RPi.GPIO was installed and everything worked beautifully.

PR #914, opened 2024-12-01
Added support for D-Robotics RDK-X3

hajimef:
Added support for D-Robotics RDK-X3
https://developer.d-robotics.cc/rdkx3


makermelissa (approved):
Thank you.

PR #921, opened 2024-12-22
Update board.py

meikuam:
Add Orange Pi 3 LTS board support, same as Pi 3 pinout


makermelissa:
Closing in favor of #952

PR #928, opened 2025-01-11
Radxa Zero 3W: add pwm support

UnicodeMan:
Add support for PWM on Radxa Zero 3 boards.
PWM8, PWM9, as the overlays for rk3566 are scarce, this works using rk3568 overlays on:
- Armbian latest
- Linux version 6.1.84-vendor-rk35xx

### DIFF
--- a/src/adafruit_blinka/microcontroller/rockchip/rk3566/pin.py
+++ b/src/adafruit_blinka/microcontroller/rockchip/rk3566/pin.py
@@ -241,6 +241,16 @@ if board in ("ODROID_M1S"):
         globals()[alias + "_TX"] = GPIO2_A4
         globals()[alias + "_RX"] = GPIO2_A3
         uartPorts.append((int(alias[-1]), GPIO2_A4, GPIO2_A3))
+        
+if board in ("RADXA_ZERO3"):
+    alias = get_pwm_chipid("fe6f0000.pwm")
+    if alias is not None:
+        globals()["PWM" + alias] = GPIO3_B1
+        pwmOuts.append(((int(alias[-1]), 0), GPIO3_B1))
+    alias = get_pwm_chipid("fe6f0010.pwm")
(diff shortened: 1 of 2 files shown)


makermelissa:
Please rebase with the latest code and push again. I updated it to use Python 3.8.

UnicodeMan:
Rebased.

UnicodeMan:
Well, code works for me fine. In any case, i will periodically rebase, until CI passes eventually, if a commit fixes it. If there is something wrong on my end, I would gladly try resolving it.

makermelissa:
Fixed formatting myself since I sat on this for so long.

src/adafruit_blinka/microcontroller/rockchip/rk3566/pin.py
```diff
@@ -242,6 +242,16 @@
         globals()[alias + "_RX"] = GPIO2_A3
         uartPorts.append((int(alias[-1]), GPIO2_A4, GPIO2_A3))
 
+if board in ("RADXA_ZERO3"):
```
ezio-melotti, line 245: A bit late to the party, but I believe this might be wrong.
Here the `()` have no effect (they don't create a tuple), so this is a substring check.  If `board` is equal to `"RADXA_ZERO3"` it will still work, but it will also work if `board` is e.g. `"RADXA"` or `"ZERO"`:
```py
>>> "RADXA_ZERO3" in ("RADXA_ZERO3")
True
>>> "RADXA" in ("RADXA_ZERO3")
True
>>> "ZERO" in ("RADXA_ZERO3")
True
```

You probably want `if board in {"RADXA_ZERO3"}:` (using sets) or `if board in ("RADXA_ZERO3",):` (using tuples).

Note that the same pattern is also found above, at line 209.

@makermelissa, let me know what you think and if you would like me to make a PR to fix this.

PR #931, opened 2025-01-20
neopixel_write: add support for PWM1 pins on Raspberry Pis

Roman3349:
Adds support for NeoPixel on PWM1 pins (13, 19, 41 and 45) on Raspberry Pis.


makermelissa (approved):
Looks good. Looking at the code it appears to use LED Channel 1 if used on the pins you mentioned, otherwise it defaults back to current functionality.

PR #936, opened 2025-01-27
Change Setup.py to use elif statements to avoid doubling requirements

makermelissa:
This changes Setup.py to use elif statements to avoid installing both Pi 4 and Pi 5 requirements on Pi 5. This may not fix the problem, but shouldn't break it either.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,10 +24,18 @@ board_reqs = []
 if os.path.exists("/proc/device-tree/compatible"):
     with open("/proc/device-tree/compatible", "rb") as f:
         compat = f.read()
+    # Jetson Nano, TX2, Xavier, etc
     if b"nvidia,tegra" in compat:
         board_reqs = ["Jetson.GPIO"]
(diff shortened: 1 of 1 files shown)


caternuson (approved):

makermelissa:
Looks like it worked!

jerryneedell:
FYI -- after installing Blinka 8.52.0 (using the script from the guide in a new venv) it seems to work OK, but then if I use pip to install a circuitptyhon library
(I tried both  adafruit-circuitpyton rfm9x and adafruit-circuitpython-ticks)
pip  reinstalls RPi.GPIO 0.7.1
and the "SOC" failure recurs.
If the them do 
pip uninstall rpi-gpio
everything work fine.

Somethting appears to still be amiss with pip and RPI.GPIO

makermelissa:
Hmm, that is odd and adds to the mystery. I guess I need to figure out where it's getting that requirement from.

PR #937, opened 2025-01-30
Add rotaryio module

makermelissa:
I decided to go with a threading/poll method as this will work more universally across boards. In the future, I may optimize this by implementing interrupts where available (such as using the raspberry pi's Rpi.GPIO package).


lennyomg:
`pip install` for Blinka 8.56.0 does not install rotaryio module. Should it be added to `setup.py`?

PR #941, opened 2025-02-11
Add PIO implementation of rotaryio for the Pi 5

makermelissa:
When I was originally tasked with adding rotaryio, this was because @jepler had requested it for the purposes of testing out the PIO implementation. I had written a generic one that worked with any linux board using threading.

This PR adds the PR implementation for the Pi 5 and falls back to the generic implementation for any other linux board. In order to test, a couple extra things need to be in place.

1. The PIO setup and permissions needs to be done as in https://learn.adafruit.com/circuitpython-on-raspberrypi-linux/using-neopixels-on-the-pi-5.
2. The adafruit_pioasm and adafruit_rp1pio libs needs to be installed with: `pip install Adafruit-Blinka-Raspberry-Pi5-Piomatter adafruit-circuitpython-pioasm`


jepler (commented):
Thank you! Looks good, didn't test.

Also, I just released the rp1pio version that allows up to 4 encoders.

jepler (approved):

PR #945, opened 2025-02-23
added support for orange pi 5 max

ErikApption:
Added support for Orange pi 5 Max which has same pin layout as orange pi 5
depends on https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/380


makermelissa (approved):
Looks good to me. Thanks.

PR #949, opened 2025-03-04
Add load_settings_toml

justmobilize:


setup.py
```diff
@@ -99,6 +99,7 @@
         "pyftdi>=0.40.0",
         "adafruit-circuitpython-typing",
         "sysv_ipc>=1.1.0;sys_platform=='linux' and platform_machine!='mips'",
+        "toml>=0.10.2;python_version<'3.11'",
```
justmobilize, line 102: `tomllib` added in 3.11

src/adafruit_blinka/__init__.py
```diff
@@ -74,6 +81,41 @@ def unlock(self):
             self._locked = False
 
 
+def load_settings_toml(*, return_toml=False):
+    """Load values from settings.toml into os.environ, so that os.getenv returns them."""
+    if not os.path.isfile("settings.toml"):
+        raise FileNotFoundError("settings.toml not cound in current directory.")
```
justmobilize, line 87: File not found

src/adafruit_blinka/__init__.py
```diff
@@ -74,6 +81,41 @@ def unlock(self):
...
+        except tomllib.TOMLDecodeError as e:
+            raise tomllib.TOMLDecodeError("Error with settings.toml file.") from e
+
+    invalid_types = set()
+    for key, value in settings.items():
+        if not isinstance(value, (bool, int, float, str)):
+            invalid_types.add(type(value).__name__)
+    if invalid_types:
+        invalid_types_string = ", ".join(invalid_types)
+        raise ValueError(
```
justmobilize, line 102: invalid types

src/adafruit_blinka/__init__.py
```diff
@@ -74,6 +81,41 @@ def unlock(self):
...
+    print("settings.toml found. Updating environment variables:")
+    with open("settings.toml", "rb") as toml_file:
+        try:
+            settings = tomllib.load(toml_file)
+        except tomllib.TOMLDecodeError as e:
+            raise tomllib.TOMLDecodeError("Error with settings.toml file.") from e
+
+    invalid_types = set()
+    for key, value in settings.items():
+        if not isinstance(value, (bool, int, float, str)):
```
justmobilize, line 98: these all get converted into strings (as they would be in the environment)

justmobilize (commented):
@dhalbert and @FoamyGuy want to take a look?

src/adafruit_blinka/__init__.py
```diff
@@ -74,6 +81,41 @@ def unlock(self):
             self._locked = False
 
 
+def load_settings_toml(*, return_toml=False):
```
FoamyGuy, line 84: @justmobilize does the wildcard `*` arg here do something specific?

src/adafruit_blinka/__init__.py
```diff
@@ -74,6 +81,41 @@ def unlock(self):
...
+
+    for key, value in settings.items():
+        key = str(key)
+        if key in os.environ:
+            print(f" - {key} already exists in environment")
+            continue
+        os.environ[key] = str(value)
+        print(f" - {key} added")
+
+    if return_toml:
```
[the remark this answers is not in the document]
FoamyGuy: I think it's okay to just return settings all the time instead of make it conditional with an argument. In the majority of cases people will have copied code that just calls the function and ignores the result.

src/adafruit_blinka/__init__.py
FoamyGuy: Ahh, tinkering in REPL a little I see it basically forces you to pass the argument by name only.

justmobilize: Yup. I find it helpful for methods like this. Makes updating the function signature easier if ever needed

src/adafruit_blinka/__init__.py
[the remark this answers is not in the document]
dhalbert: I agree, no reason to control whether or not to return, removes an arg, and reduces code size. `settings` will get gc'd the same way regardless of whether or not it is returned, if the caller does not assign it to anything.

dhalbert (changes requested):

src/adafruit_blinka/__init__.py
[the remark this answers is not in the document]
justmobilize: Sounds good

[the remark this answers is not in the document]
justmobilize: Fixed here: 046217cbbf14ba5a595d9423f58a7961f0ab51cd

FoamyGuy (approved):
Thank you, this looks good to me. Tested successfully in the REPL on a pi5

PR #950, opened 2025-03-04
update minimum python version to 3.9

FoamyGuy:
cc @makermelissa we discussed minimum version for Blinka and Circuitpython_Typing in the weeds during Monday's meeting a bit.  Am definitely open to input from you as well on updating it, and if you know whether there is anywhere else in this repo that it might need to be changed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ setup(
     long_description_content_type="text/x-rst",
     author="Adafruit Industries",
     author_email="circuitpython@adafruit.com",
-    python_requires=">=3.7.0",
+    python_requires=">=3.9.0",
     url="https://github.com/adafruit/Adafruit_Blinka",
     package_dir={"": "src"},
     packages=find_packages("src") + ["micropython-stubs"],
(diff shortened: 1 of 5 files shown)


makermelissa:
One of the things holding us back on updating to a more recent version of python is the Jetson Nano, which last time I checked would only install up to version 3.8. So before we merge this I need to check a few things:

1. Update the Jetson Nano and see if Python > 3.8 is available (internet is currently down, so will need to wait until tomorrow unless I figure out an alternative sooner)
2. If not, there may be an alternate way such as building from sources, so a guide section update would need to be done in such a case.

makermelissa:
Ok, I figured out a good alternative and will give it a shot.

makermelissa:
Python 3.8 is still the latest through `sudo apt install`. There may be another way to update and I'll look into that.

setup.py
```diff
@@ -60,7 +60,7 @@
     long_description_content_type="text/x-rst",
     author="Adafruit Industries",
     author_email="circuitpython@adafruit.com",
-    python_requires=">=3.7.0",
+    python_requires=">=3.9.0",
```
makermelissa, line 63: I'd like to keep this at 3.8.0 so that we don't leave the NVIDIA jetson folks behind.

makermelissa (changes requested):
I don't mind bumping the testing to a newer version, but would like to leave the setup.py to require no newer than 3.8.0 because that would remove usability from a bunch of boards.

makermelissa:
Closing as per explanation above.

Issue #966, opened 2025-05-01
lgpio_pin.py fails looking for driver after pi 5 debian update with external SSD


pwagsc:
### Board Name

PI 5

### Steps

This was posted in PI forum:  My failure was identical and occurred after upgrade to 6.12.20+rpt-rpi-2712 on my PI 5. Peering at lgpio_pin.py code reveals that it fails when it tries to read the symbolic linked compatible file for gpiochip14 since the symlink does not exist. In fact gpiochip14 didn't exist before update. So the failure is the absent symlink and compatible file. All other gpiochipxx directories have the symlink and file. It is above my pay grade to determine the flaw that allowed creation of a new gpiochip that has no symlink. Some debian expert will have to sort that out. If it is legal to have no symlink then adafruit needs to fix their code (or for robustness). A simple try ... except pass around lines 18-26 and all is back to working again as it then ignores any gpiochip that doesn't have a compatible file/symlink. As a note, the .glob of files to be inspected just happened to start with gpiochip14 so it couldn't find chip0.
As a side note, I see the list of drivers doesn't include raspberrypi,bcm2712-gpio. However on my system driver 0 has raspberrypi,rp1-gpio in compatible file. So don't know what is going on there but it doesn't hurt anything at present.

### Description

Only happens after upgrade to 6.12.20+rpt-rpi-2712 on my PI 5

makermelissa:
Thanks, I'll test it out and see if I can reproduce the error.

makermelissa:
For reference, it appears your post is here: https://forums.raspberrypi.com/viewtopic.php?p=2313555&hilit=gpiochip14#p2312930

Do you have some test code you were running.

makermelissa:
Tested this out with 6.12.25+rpt-rpi-2712 (current latest) and did not run into these issues. Perhaps it was a bug in 6.12.20+rpt-rpi-2712 and they quickly released it again. Please update the Pi and give it another try. Also, to note, I think you meant `gpiochip4` rather than `gpiochip14`.

pwagsc:
Hi Melissa,

[...]

PR #967, opened 2025-05-15
fix compatibility with differing `/sys/bus/gpio/devices/gpiochip*`

mrossington:
When using a USB to Serial cable I note that initialising the blinka library would crash. I'd receive an error saying that it cannot open the file `/sys/bus/gpio/devices/gpiochip14/of_node/compatible`. This was because the FTDI cable had added an additional gpioChip which didn't have the expected path structure. 

I feel like it's safe to add an `is_file()` check to this path before opening and this does make the issue go away on my setup.

Please let me know if there are:

 - Coding standards I may have missed
 - Tests that I need to run and ensure I haven't broken
 - A better way to handle this

Additionally, is it worth looking if we do this sort of blind read elsewhere and fix those too? Could then add a generic function.


makermelissa (approved):
Sounds good. Hopefully this fixes #966.

PR #972, opened 2025-06-18
Convert Blinka to use pyproject.toml

makermelissa:
Using setup.py is a bit outdated. The main reason Blinka has held out for so long was because of the nature of setup.py checking processor to determine which dependencies to install. This was especially important when the Pi 5 relied on rpi-lgpio (which it no longer does). Now that that has been sorted out, it seems appropriate to convert to pyproject.toml.

I think this will solve many of the installation problems folks have had.


FoamyGuy (approved):
Looks good to me. Thank you

Issue #975, opened 2025-06-19
import digitalio fails under latest releases


FoamyGuy:
### Board Name

Ubuntu and actions containers

### Steps

Install adafruit-blinka and open repl and run `import digitalio`

### Description

outputs this error:
```
>>> import digitalio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/timc/repos/circuitpython/.cpy_venv/lib/python3.12/site-packages/digitalio.py", line 12, in <module>
    from adafruit_blinka.agnostic import board_id, detector
ModuleNotFoundError: No module named 'adafruit_blinka'
```

jscheidtmann:
Same problem here on a freshly installed Raspi 5 running latest Raspian with fresh python virtual environment.

I get the same error already on `import board`, when running the usage example.

caternuson:
This issue may be more generic. Seeing the same error on a Pi5 after a doing a fresh blinka install.

```
pi@raspberrypi:~ $  source env/bin/activate
(env) pi@raspberrypi:~ $ python3
Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import board
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pi/env/lib/python3.11/site-packages/board.py", line 21, in <module>
    from adafruit_blinka.agnostic import board_id, detector
ModuleNotFoundError: No module named 'adafruit_blinka'
```

See this forum post for more details:
https://forums.adafruit.com/viewtopic.php?t=218842

jscheidtmann:
Good morning. A `...venv/bin/pip uninstall Adafruit-Blinka` followed by a `pip install Adafruit-Blinka` installed 8.60.3. The bug did not show up with this version and the service could start with-out problems.

FoamyGuy:
Thank you!

jamb1955:
Thank you for the support !
I restarted the process from scratch, some more careful use of the VENV idea and I got it ALL to work!
I can control the display and use it just as I did on the Pico, but with this library.
GREAT thank you all!

PR #976, opened 2025-06-19
include adafruit_blinka package in build

FoamyGuy:
I think this resolves: #975 

with this change I tested locally by doing `pip install .` and then going to the REPL and I can now successfully run `import digitalio` 

All testing was ubuntu 24.04 under python 3.12

Perhaps this makes the items that follow it in the packages list redundant though? Maybe those could be removed, I didn't try it, and they look potentially hardware dependent so I'm not sure how to test if we do remove them.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ py-modules = [
 [tool.setuptools.packages.find]
 where = ["src"]
 include = [
+    "adafruit_blinka",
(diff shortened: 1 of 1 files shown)


tekktrik (approved):
Based on some testing I did with package building, I believe you still need the other packages listed.  I think it would be different if you included `"adafruit_blinka*"` since that would get all paths, but as is this is good.

PR #981, opened 2025-06-23
Revert back to using setup.py to fix install problems

makermelissa:
The library was recently converted to use pyproject.toml, which isn't flexible enough to deal with determining specific processor package requirements. Installing them anyways results in conflicting packages. This undos those changes.


dhalbert (approved):
I did not test but assuming you have, fine with me to merge and release.

makermelissa:
Not so easy to test without releasing, but since this puts it in the state it was in prior to the pyproject.toml changes, it should be fine.

PR #983, opened 2025-06-23
Move tomllib imports to function so it works on MicroPython

### LINKED ISSUE #962: no module named toml when importing in MicroPython
### Board Name

Raspberry Pi Pico

### Steps

I was following this guide on using CircuitPython libraries in MicroPython and when I tried to use it, I got an import error for the toml module.

I've cloned the main toml module and ripped out all the bits that don't work with MicroPython and ended up with a version that's a bit janky but does seem to work (https://github.com/benevpi/toml-micropython).

Before tidying this up and making it a bit nicer, I wanted to check if this was the right way to solve this problem or if I've done something wrong to get here.

### Description

_No response_

### Additional information

_No response_

makermelissa:
This should fix #962.

### DIFF
--- a/src/adafruit_blinka/__init__.py
+++ b/src/adafruit_blinka/__init__.py
@@ -10,11 +10,6 @@
 
 import os
 
-try:
-    import tomllib
-except ImportError:
-    import toml as tomllib
-
 
 class Enum:
     """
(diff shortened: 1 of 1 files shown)


makermelissa:
Eventually, it would be nice to add support for https://github.com/BrianPugh/micropython-tomli or something similar.

tekktrik (approved):
Ahh, nice catch!

PR #987, opened 2025-07-02
Improve undetected I2C and SPI messages

makermelissa:
When I2C or SPI is not enabled, the messages were telling the user that the hardware was not detected. This adds a hint about the most obvious thing to check.


dhalbert (approved):
Good idea!

PR #989, opened 2025-07-05
Detect gpiochip via libgpiod for H616

mlodedrwale:
Use python3-libgpiod to scan /dev/gpiochip* and match each chip’s label, replacing the sysfs-only lookup. This ensures reliable detection on H616 and adds support for Orange Pi SBCs where the old method failed.


makermelissa (changes requested):
This is depending upon a specific version of GPIOD. Some versions have `chip` and some have `Chip`. It would be better to make use of https://github.com/adafruit/Adafruit_Blinka/blob/main/src/adafruit_blinka/microcontroller/generic_linux/libgpiod_chip.py which will figure out the appropriate class by looking at specific traits and returning the correct one as Chip so it's consistent.

makermelissa:
Closing due to lack of response about requested changes.

PR #1001, opened 2025-08-15
Support for Particle Tachyon

technobly:
- Adds Blinka support for [Particle Tachyon](https://www.particle.io/tachyon/)


ladyada:
@makermelissa did you check the changes to libgpio.py code? @technobly why were those changes made?

technobly:
@ladyada IIRC it was erroring when I tried running with libgpiod v2.1.3
If you don't have an easy way to check it, you can revert it because currently Tachyon uses v1.4.1 of libgpiod on ubuntu v20.04.  v2.x should be supported in ubuntu 24.04 when that is released for Tachyon, so we'll need it soon enough though.

makermelissa:
I reverted the libgpiod changes in #1003. @technobly, if you find an error, please open up a new issue in the future. Thanks.

PR #1033, opened 2025-12-27
Add banana pi p2 pro

dmlb2000:
Add banana pi p2 pro to the board list.

https://wiki.banana-pi.org/Banana_Pi_BPI-P2_Pro


dmlb2000:
Okay so this depends on https://github.com/adafruit/Adafruit_Python_PlatformDetect/pull/404 first...

PR #1054, opened 2026-04-03
Change New Board Request to start new discussion

### LINKED ISSUE #1007: Move new board requests into a discussion
In order to reduce the issues from being hidden amongst new board requests, they should be moved into a discussion. We will also need to put in place a way to direct users to add to the discussion rather than opening an issue.

makermelissa:
Fixes #1007. In order to be able to make it so the Blinka issues are more manageable, we had previously discussed moving new board requests to discussions so that actual bugs and features could be better seen. As it is now, more than %0% or the issues are for new board requests. After this change, I will start moving the issues over (or if possible, have AI do it).


makermelissa:
I started migrating them anyways. This really only applies to new issues.

makermelissa:
Merging since it's more of a GitHub configuration change, I already tested, and we already have new board requests coming in.

PR #1104, opened 2026-04-15
Add in_waiting property to busio.UART

makermelissa-piclaw:
## Summary

Adds the `in_waiting` property to `busio.UART` to match CircuitPython's API. Returns the number of bytes available in the input buffer.

This is needed by libraries like [Adafruit_CircuitPython_GPS](https://github.com/adafruit/Adafruit_CircuitPython_GPS) that rely on `in_waiting` to check for available data before reading.

### Changes

| File | What changed |
|------|-------------|
| `src/busio.py` | Added `in_waiting` property with fallback chain: custom `in_waiting` → MicroPython `any()` → `NotImplementedError` |
| `src/adafruit_blinka/microcontroller/rp2040/uart.py` | Added `in_waiting` property wrapping MicroPython's `uart.any()` |

### How it works

- **RP2040** (and future custom UART classes): Uses the `in_waiting` property on the underlying UART implementation
- **Generic MicroPython boards**: Falls back to `uart.any()` which is MicroPython's equivalent
- **Unsupported platforms**: Raises `NotImplementedError` with a clear message

Fixes #1046

### DIFF
--- a/src/adafruit_blinka/microcontroller/rp2040/uart.py
+++ b/src/adafruit_blinka/microcontroller/rp2040/uart.py
@@ -50,3 +50,8 @@ class UART:
     def write(self, buf):
         """Write to the UART from a buffer"""
         return self._uart.write(buf)
+
+    @property
+    def in_waiting(self):
+        """The number of bytes in the input buffer, available to be read"""
+        return self._uart.any()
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Looks good.

Fancy2209:
When I tested it, return self._uart.any() wouldn't work on the GPS Lib
I had to do
```py
result = self._uart.any()
return result if result == 0 else int(result / 8)
```
However not sure if the magic 8 is RPi Pico specific

PR #1109, opened 2026-04-22
Ensure libgpiod resources are cleaned up on exit

### LINKED ISSUE #435: Make sure to kill libgpiod threads
Not sure what a good answer to this is, or if it's even something that should be required of Blinka. But starting here.

This issue is a good example of what can happen:
https://github.com/adafruit/Adafruit_CircuitPython_DHT/issues/63
and more info in a forum related post here:
https://forums.adafruit.com/viewtopic.php?f=60&t=175209#p859042
which I'm copying here:

I think the general issue here is that these codes are all infinite loops that are being aborted using <CTRL><C>. As a result, the DHT exit() never gets called:
https://github.com/adafruit/Adafruit_CircuitPython_DHT/blob/ceb883d09587612d2f96037f54c93805e4175db3/adafruit_dht.py#L59
and so the pulseio deinit() does not get called:
https://github.com/adafruit/Adafruit_Blinka/blob/bc28caa0fe210741bda9586a134adad3794bda5e/src/adafruit_blinka/microcontroller/bcm283x/pulseio/PulseIn.py#L99
and so the background process handling the pulsein does not get terminated. So when the code is run a second time, that still running process still has the pin taken.

Further, the current example has some try/except logic:
https://github.com/adafruit/Adafruit_CircuitPython_DHT/blob/master/examples/dht_simpletest.py
but this will not catch a <CTRL><C> event since KeyboardInterrupt inherits from BaseException, not Exception:
https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt

makermelissa-piclaw:
## Summary

Fixes #435 — GPIO lines held by libgpiod and PulseIn subprocesses were not being properly released when scripts were interrupted (Ctrl+C, SIGTERM), causing "Unable to set line to input" errors on subsequent runs.

## Problem

When a Python script using Blinka's GPIO or PulseIn is interrupted:

1. **libgpiod 1.x Pin** had no `__del__` or `deinit()` method, so GPIO lines were never released unless `init()` was called again to reconfigure
2. **libgpiod Chip** objects (both 1.x and 2.x) were never explicitly closed
3. **PulseIn subprocesses** relied solely on `atexit` handlers which don't fire on SIGTERM, and the `deinit()` method would crash if called twice
4. **`digitalio.DigitalInOut.deinit()`** just did `del self._pin` which doesn't release GPIO lines on libgpiod 1.x

## Changes

- **libgpiod Pin (1.x & 2.x):** Add `deinit()` method to explicitly release lines and close chips. Add `__del__` as a GC safety net.
- **libgpiod Chip (1.x & 2.x):** Add `__del__` to close chip handles on garbage collection.
- **PulseIn (bcm283x, meson_g12, a311d):** Add `__del__` safety net. Make `deinit()` idempotent (safe to call multiple times). Add SIGTERM signal handler alongside existing `atexit` handler to ensure subprocesses and message queues are cleaned up.
- **digitalio.DigitalInOut:** `deinit()` now calls `pin.deinit()` when available instead of just deleting the reference.

### DIFF
--- a/src/adafruit_blinka/microcontroller/amlogic/a311d/pulseio/PulseIn.py
+++ b/src/adafruit_blinka/microcontroller/amlogic/a311d/pulseio/PulseIn.py
@@ -7,6 +7,7 @@ import subprocess
 import os
 import atexit
 import random
+import signal
 import struct
 import sysv_ipc
 
@@ -29,6 +30,18 @@ def final():
 
 
 atexit.register(final)
+
+
+def _signal_handler(signum, frame):  # pylint: disable=unused-argument
+    """Handle SIGTERM/SIGINT to ensure cleanup runs"""
+    final()
+    raise SystemExit(1)
+
+
+try:
+    signal.signal(signal.SIGTERM, _signal_handler)
+except (OSError, ValueError):
+    pass  # Not all environments allow signal handling
 
 
 # pylint: disable=c-extension-no-member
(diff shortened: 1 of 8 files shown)


makermelissa (approved):
Looks good.

PR #1117, opened 2026-07-07
Add PlatformDetect dependency update workflow

makermelissa:
Adds a scheduled/manual workflow that updates Blinka's minimum Adafruit-PlatformDetect dependency only when PyPI has a newer release. If the dependency is already current, the workflow exits without opening a new PR.


makermelissa:
Closing this PR because the branch was pushed directly to the upstream Adafruit repository by mistake. I am reopening from the makermelissa fork instead.

PR #1124, opened 2026-09-06
Ship micropython as a typed package, drop micropython-stubs

### LINKED ISSUE #11221: Same stub file provided by multiple circuitpython packages
### CircuitPython versions

```text
firmware:
circuitpython == 10.3.0
bootloader == 4.0.0

python dependencies:
├── adafruit-circuitpython-logging v5.6.4
│   ├── adafruit-blinka v9.2.0
│   └── typing-extensions v4.16.0
├── circuitpython-stubs v10.3.0
```

### Description

I'm using `circuitpython-stubs` and `adafruit-circuitpython-logging` in my project. An experimental feature in `uv sync` recently showed this warning:

```sh
UV_MALWARE_CHECK=1 uv sync
warning: The file `micropython-stubs/__init__.pyi` is provided by more than 
         one package, which causes an install race condition and can result
         in a broken module. Packages containing the file:
* adafruit-blinka (adafruit_blinka-9.2.0-py3-none-any.whl)
* circuitpython-stubs (circuitpython_stubs-10.2.1-py3-none-any.whl)
```

They are stubs (not run-time) and I've not experienced any type-checking issues so far. But the two `__init__.py` are different, so there is a possibility.

4RH1T3CT0R7:
Since adafruit/circuitpython#10485, `circuitpython-stubs` ships its own `micropython-stubs/__init__.pyi`, so installing it next to Blinka writes the same file twice. pip silently keeps whichever wheel lands last, and uv now warns about the install race (adafruit/circuitpython#11221).

Blinka owns the runtime `micropython` module, so the PEP 561 way to type it is inline rather than a separate stub-only package. This moves `micropython.py` to `micropython/__init__.py`, adds the annotations that lived in the stub plus a `py.typed` marker, and drops the `micropython-stubs` package from setup.py. Behaviour is unchanged. Checked in a clean venv that mypy resolves `micropython.const` from the package alone, and that with `circuitpython-stubs` installed the stubs take precedence and the two wheels no longer share any file.

Fixes adafruit/circuitpython#11221


alec-bike:
I tested this branch in my project and it resolves the issue seen in https://github.com/adafruit/circuitpython/issues/11221.

4RH1T3CT0R7:
Thanks for testing it, @alec-bike. On the review note: the NoReturn annotations on viper and asm_thumb are copied unchanged from the micropython-stubs file this replaces, and both decorators raise SyntaxError at runtime, so the typing stays exactly as it was before.

